PR #3, opened 2024-01-22
Add `DESTINATION` envvar

fluiderson:


src/server.rs
```diff
@@ -72,12 +73,35 @@ pub(crate) async fn new(
...
+) -> Json<Response> {
+    let wss = database.rpc().to_string();
+    let mul = database.properties().await.decimals;
+
+    match abcd(database, Some(recipient), order, price).await {
+        Ok(re) => Response::Success(re),
+        Err(error) => Response::Error(Error {
+            wss,
+            mul,
+            version: env!("CARGO_PKG_VERSION").into(),
```
Slesarew, line 88: Compiled binary does not know this value. Please confirm that this is an issue and we'll add another versioning method later.

start.sh
```diff
@@ -2,4 +2,5 @@ KALATORI_HOST="127.0.0.1:16726" \
 KALATORI_SEED="" \
 KALATORI_RPC="ws://127.0.0.1:9944" \
 KALATORI_DECIMALS="12" \
+KALATORI_DESTINATION="0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d" \
```
Slesarew, line 5: is this a well-known account? Good practices require us to post public key or full derivation for this lest we accidentally send something there and it will get stuck.

Slesarew (approved):

PR #12, opened 2024-03-03
ci: init

Slesarew:
usual stuff we'll put into mandatory checks as always

### DIFF
--- a/.github/workflows/rust-clippy.yml
+++ b/.github/workflows/rust-clippy.yml
@@ -1,0 +1,38 @@
+name:                         Rust clippy
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - stable
+
+jobs:
+  linter:
+    name:                     Cargo clippy
+    runs-on:                  ubuntu-latest
+    steps:
+
+      - name:                 Cancel Previous Runs
+        uses:                 styfle/cancel-workflow-action@0.12.1
+        with:
+          access_token:       ${{ github.token }}
+
+      - name:                 Checkout sources
+        uses:                 actions/checkout@v4.1.1
+        with:
+          fetch-depth:        50
+          submodules:         'recursive'
+
+      - name:                 Install Rust stable toolchain
+        uses:                 actions-rs/toolchain@v1.0.7
+        with:
+          profile:            minimal
+          toolchain:          stable
+          override:           true
+
+      - name:                 Rust Cache
+        uses:                 Swatinem/rust-cache@v2.7.3
+
+      - name:                 cargo clippy
+        run:                  cargo clippy --all-targets -- -D warnings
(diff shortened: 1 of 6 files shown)


kirushik:
Can we please also have an additional check that you can't merge into `main` without bumping at least the patch version?

As per request by @lleokaganov, seems super-useful for debugging

Slesarew:
@kirushik This would not work for building off-main branch, which is probably even more problematic case, maybe we just concat git commit hash to `health` response and everywhere else where this is relevant in github builds until we start putting this on crates?

Slesarew:
https://crates.io/crates/git-version maybe

kirushik:
> This would not work for building off-main branch, which is probably even more problematic case

Well, it's possible to have a check which only triggers when PR is against a main branch; I probably don't understand the problem?

In my mind something like this (CAUTION: I haven't tested it) should work:

```yaml
on:
 pull_request:
    branches: [ "main" ]

jobs:
 check-version-bump:
    runs-on: ubuntu-latest
    steps:
    - name: Checkout PR branch
      uses: actions/checkout@v2
      with:
        ref: ${{ GITHUB_HEAD_REF }}

    - name: Checkout main branch
      uses: actions/checkout@v2
      with:
        ref: ${{ GITHUB_BASE_REF }}
        path: main-branch

    - name: Check version bump
      run: |
        PR_VERSION=$(grep '^version' Cargo.toml | cut -d'"' -f2)
        MAIN_VERSION=$(grep '^version' main-branch/Cargo.toml | cut -d'"' -f2)
        if [ "$PR_VERSION" == "$MAIN_VERSION" ]; then
          echo "Error: Version has not been bumped."
          exit 1
        fi
```

.github/workflows/rust-clippy.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          profile:            minimal
+          toolchain:          stable
+          override:           true
+
+      - name:                 Rust Cache
+        uses:                 Swatinem/rust-cache@v2.7.3
+
+      - name:                 cargo clippy
+        run:                  cargo clippy --all-targets -- -D warnings
```
kirushik, line 38: If you're doing `-D warnings` anyway, why not do the full package:

```suggestion
        run:                  cargo clippy -- -D warnings -D clippy::pedantic -D clippy::correctness -D clippy::complexity -D clippy::perf
```

.github/workflows/rust-clippy.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          profile:            minimal
+          toolchain:          stable
+          override:           true
+
+      - name:                 Rust Cache
+        uses:                 Swatinem/rust-cache@v2.7.3
+
+      - name:                 cargo clippy
+        run:                  cargo clippy -- -D warnings -D clippy::pedantic -D clippy::correctness -D clippy::complexity -D clippy::perf
```
Slesarew, line 38: important thing we forgot!
```suggestion
        run:                  cargo clippy --all-targets -- -D warnings -D clippy::pedantic -D clippy::correctness -D clippy::complexity -D clippy::perf
```

kirushik (approved):

PR #14, opened 2024-03-04
chore: fixed all the minor Clippy rants

kirushik:
As a preparation for #12 being merged

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,6 @@ shadow_reuse = "warn"
 shadow_same = "warn"
 shadow_unrelated = "warn"
 cargo_common_metadata = "warn"
+
+[package.metadata.scripts]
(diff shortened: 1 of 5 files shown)


kirushik:
I still find https://github.com/Alzymologist/Kalatori-backend/blob/5ba7f4db5faad3baff183058eaf88ee6856a2c0b/src/database.rs#L316-L325 and https://github.com/Alzymologist/Kalatori-backend/blob/5ba7f4db5faad3baff183058eaf88ee6856a2c0b/src/rpc.rs#L466-L488 suspicious, and would've preferred this would be a more concise logic.

Consumer asked for `invoices()`? Give them an `Iterator` over all the properly parsed elements in the db.

Slesarew:
we can merge #12 first to make sure this passes; branch protection rules are not set yet.

Slesarew (approved):

Slesarew:
Is this redundant with #18?

Slesarew:
ok it is deprecated I suppose

PR #17, opened 2024-03-05
ci: Added build and draft release job

Nyoxis:
For `<tag>` used package version, results in duplicate releases with the same name, if version wasn't bumped beforehand.


Nyoxis:
> For `<tag>` used package version, results in duplicate releases with the same name, if version wasn't bumped beforehand.

Added package version bump check, excluding this situation.

On pull request event, head and base branches are compared. On push event, this check compares current commit with the previous commit by SHA, thus resulting in comparing versions when `commit --amend`.

kirushik (commented):
In principle, I'd rather have a `is_version_greater.sh` script which exits with a `0` exit code if it is indeed greater, and with a non-zero one if it's not. (Rather than putting those into the STDOUT and then comparing the values down the road.)

If it would be easy for you to fix — please do; if it will take more than, say, 30 minutes — let's add an issue to track this and merge this as it is.

PR #18, opened 2024-03-05
feat: AssetHub support

fluiderson:


kirushik (dismissed):
LGTM from the feature-set posisions.
I'll defer to @Slesarew to pass the final, engineering judgement.

kalatori-ah/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    let (database, last_saved_block) = Database::initialise(
+        database_path,
+        override_rpc,
+        pair,
+        endpoint_properties,
+        destination,
+    )
+    .context("failed to initialise the database module")?;
+
+    let processor = Processor::new(api_config, database.clone(), shutdown_notification.clone())
```
Slesarew, line 254: >`database(clone)`

Right now it seems that this works OK concurrently, but later it would be best to move db handler into separate thread and talk to it through messages to isolate it (and make it even faster by disk write yelds).

I would think that in times of refactor this is one of the highest priority things.

kalatori-ah/src/rpc.rs
```diff
@@ -0,0 +1,1352 @@
...
+}
+
+async fn fetch_runtime(
+    methods: &LegacyRpcMethods<RuntimeConfig>,
+    backend: &impl Backend<RuntimeConfig>,
+) -> Result<(Metadata, RuntimeVersion)> {
+    let best_block = fetch_best_block(methods).await?;
+
+    Ok((
+        fetch_metadata(backend, best_block)
```
Slesarew, line 68: shouldn't this be finalized block?

kalatori-ah/src/rpc.rs
```diff
@@ -0,0 +1,1352 @@
...
+                let transfers = vec![construct_transfer(
+                    &changes.invoice.recipient,
+                    price - EXPECTED_USDX_FEE,
+                    self.database.properties().await.usd_asset,
+                )];
+                let tx = self
+                    .batch_transfer(current_nonce, block_hash_count, &signer, transfers.clone())
+                    .await?;
+
+                // if let Some(a) = (fee + properties.existential_deposit + price).checked_sub(balance)
```
Slesarew, line 1190: maybe this stuff should be removed, not commented out - we do have version control here

Slesarew (approved):

kalatori-ah/src/rpc.rs
fluiderson: `fetch_runtime` is used only with the API client that requires to be in sync with the best chain, not the finalized one.

PR #31, opened 2024-04-01
New API

fluiderson:


src/server.rs
```diff
@@ -1,209 +1,426 @@
...
+            description: ServerInfo {
+                version: env!("CARGO_PKG_VERSION"),
+                instance_id: String::new(),
+                debug: state.0.debug,
+                kalatori_remark: state.0.remark.clone(),
+            },
+            supported_currencies: vec![CurrencyInfo {
+                currency: "USDC".into(),
+                chain_name: "assethub-polkadot".into(),
+                kind: TokenKind::Assets,
```
kirushik, line 418: This yields `assets` instead of a singular `asset`, which violates the API defined in https://alzymologist.github.io/kalatori-api/

src/server.rs
```diff
@@ -1,209 +1,426 @@
...
-    #[allow(clippy::cast_precision_loss)]
-    let mul = 10u128.pow(dec.try_into()?) as f64;
+async fn order(
+    state: extract::State<Arc<State>>,
+    matched_path: MatchedPath,
+    path_result: Result<RawPathParams, RawPathParamsRejection>,
+    query: Query<HashMap<String, String>>,
+) -> Response {
+    match process_order(state, &matched_path, path_result, &query).await {
+        Ok((order_status, order_success)) => match order_success {
```
kirushik, line 357: This doesn't work as specified in the API docs.
It **should** return the full details about the current status of the order (including amount, currency info etc), while the only thing I'm getting is

```json
{
    "message": "",
    "order": "123456",
    "payment_status": "unknown",
    "recipient": "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",
    "server_info": {
        "debug": true,
        "instance_id": "",
        "version": "0.2.0-rc1"
    }
}
```

kirushik (changes requested):

src/main.rs
```diff
@@ -1,5 +1,478 @@
...
+    task_tracker.spawn(
+        "the shutdown listener",
+        shutdown_listener(shutdown_notification.clone()),
+    );
+
+    let (chains, currencies) = rpc::prepare(config.chain, config.account_lifetime, config.depth)
+        .await
+        .context("failed while preparing the RPC module")?;
+
+    let rpc = env::var("KALATORI_RPC").unwrap();
```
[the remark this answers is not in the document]
fluiderson: I added it just for this mockup to process a single RPC endpoint. It won't be in the final version, all RPC endpoints will be obtained only from the config.

PR #45, opened 2024-05-14
major: api_v2 and major rehaul of everything

Slesarew:
Now this is architecture framework that should work like something solid. This requires A LOT of effort at the moment.

- [x] Consider replacing DB back to redb (only if you can do it without compromising safety through redb's `static` everywhere)
- [x] replace numerous unwrap placeholders with properly listed errors
- [ ] fulfill todos in actual api2 responses
- [x] add timestamps to stored orders
- [ ] add versioning and migration stub to database
- [x] make payout engine a separate thread(s), make it actually check what it is doing and store transaction data as per `api_v2` specs

and more, but let's start from these


start.sh
```diff
@@ -1,6 +1,9 @@
 KALATORI_HOST="127.0.0.1:16726" \
-KALATORI_SEED="" \
-KALATORI_DECIMALS="12" \
-KALATORI_DESTINATION="0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d" \
```
Vovke, line 4: Update README.md to have RECEIPIENT instead of destination

src/main.rs
```diff
@@ -1,5 +1,412 @@
...
+        unparsed_host
+            .parse()
+            .map_err(|_| Error::ConfigParse("host to define a socket address".to_string()))?
+    } else {
+        DEFAULT_SOCKET
+    };
+
+    let debug = config.debug;
+
+    let database_path = 'database: {
```
Vovke, line 80: looks like the database won't ever come from env KALATORI_DATABASE

src/definitions.rs
```diff
@@ -0,0 +1,324 @@
...
+#[cfg(test)]
+#[test]
+#[allow(
+    clippy::inconsistent_digit_grouping,
+    clippy::unreadable_literal,
+    clippy::float_cmp
+)]
+
+fn balance_insufficient_precision() {
+    const DECIMALS: Decimals = 10;
```
Vovke, line 317: const DECIMALS: api_v2::Decimals = 10;

Vovke, line 317:
```suggestion
    const DECIMALS: api_v2::Decimals = 10;
```

src/chain/rpc.rs
```diff
@@ -0,0 +1,756 @@
...
+// Runtime APIs
+
+const AURA: &str = "AuraApi";
+
+pub async fn subscribe_blocks(client: &WsClient) -> Result<Subscription<BlockHead>, ErrorChain> {
+    Ok(client
+            .subscribe(
+                "chain_subscribeFinalizedHeads",
+                rpc_params![],
+                "unsubscribe blocks",
```
Vovke, line 73:
```suggestion
                "chain_unsubscribeFinalizedHeads",
```

start.sh
```diff
@@ -1,6 +1,9 @@
 KALATORI_HOST="127.0.0.1:16726" \
```
Vovke, line 1:
```suggestion
KALATORI_CONFIG="configs/chopsticks.toml" \
KALATORI_HOST="127.0.0.1:16726" \
```

start.sh
```diff
@@ -1,6 +1,9 @@
...
-KALATORI_SEED="" \
-KALATORI_DECIMALS="12" \
-KALATORI_DESTINATION="0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d" \
-KALATORI_LOG="subxt::events::events_type=off" \
+KALATORI_SEED="bottom drive obey lake curtain smoke basket hold race lonely fit walk" \
+KALATORI_DATABASE="database-ah-usdc.redb" \
+KALATORI_RPC="wss://polkadot-asset-hub-rpc.polkadot.io" \
```
Vovke, line 4:
```suggestion
KALATORI_RPC="ws://localhost:8000" \
```

Cargo.toml
```diff
@@ -10,22 +10,35 @@ readme = true
...
-env_logger = "0.11"
-log = "0.4"
-subxt = { version = "0.34", features = ["substrate-compat"] }
-axum = "0.7"
-serde = "1"
-redb = "1"
-serde_json = "1"
-hex = "0.4"
-reconnecting-jsonrpsee-ws-client = { version = "0.3", features = ["subxt"] }
+
```
Vovke, line 22:
```suggestion
reqwest = { version = "0.11", features = ["json"] }

```

src/chain/rpc.rs
```diff
@@ -0,0 +1,724 @@
...
+        _ => return Err(ErrorChain::BlockHashFormat),
+    }
+}
+
+/// fetch metadata at known block
+pub async fn metadata(
+    client: &WsClient,
+    block: &BlockHash,
+) -> Result<RuntimeMetadataV15, ErrorChain> {
+    let metadata_request: Value = client
```
Vovke, line 156:
```suggestion
    let block_str = format!("0x{}", block.to_string());
    let metadata_request: Value = client
```

src/chain/rpc.rs
```diff
@@ -0,0 +1,724 @@
...
+        .request(
+            "state_call",
+            rpc_params![
+                "Metadata_metadata_at_version",
+                "0f000000",
```
Vovke, line 161:
```suggestion
                "0x0f000000",
```

src/chain/rpc.rs
```diff
@@ -0,0 +1,724 @@
...
+                &block.to_string()
```
Vovke, line 162:
```suggestion
                &block_str
```

src/chain/rpc.rs
```diff
@@ -0,0 +1,724 @@
...
+
+pub async fn get_keys_from_storage(
+    client: &WsClient,
+    prefix: &str,
+    storage_name: &str,
+    block: &BlockHash,
+) -> Result<Value, ErrorChain> {
+    let keys: Value = client
+        .request(
+            "state_getKeys",
```
Vovke, line 100: state_getKeys got deprecated, suggested to use (state_getKeysPaged](https://polkadot.js.org/docs/substrate/rpc/#getkeyskey-storagekey-at-blockhash-vecstoragekey) instead

following implementation works, but can be optimised
```
pub async fn get_keys_from_storage(
    client: &WsClient,
    prefix: &str,
    storage_name: &str,
    block: &BlockHash,
) -> Result<Value, ErrorChain> {
    let storage_key_prefix = format!(
        "0x{}{}",
        hex::encode(twox_128(prefix.as_bytes())),
        hex::encode(twox_128(storage_name.as_bytes()))
    );

    let count = 100; // I expect it to always work, but will it actually?
    let start_key: Option<&str> = None; // Start from the beginning

    let mut params = vec![
        serde_json::to_value(storage_key_prefix.clone()).unwrap(),
        serde_json::to_value(count).unwrap(),
    ];

    if let Some(start_key) = start_key {
        params.push(serde_json::to_value(start_key).unwrap());
    }

    params.push(serde_json::to_value(block.to_string()).unwrap());

    let keys: Value = client
        .request(
            "state_getKeysPaged",
            params
        )
        .await
        .map_err(ErrorChain::Client)?;

    Ok(keys)
}
```

start.sh
Slesarew: The funny thing is, without chopsticks setup it just runs. With chopsticks it requires all that messy setup that won't even be in prod. Maybe we want several start scripts here?

src/chain/rpc.rs
Slesarew: no, I've made a special interface for it that does all dangerous conversions! If we want to always append `0x`, we should do it there. Is it a real recommendation now? Because upstream tends to break these things as we know.

shoot.sh
```diff
@@ -1 +1 @@
-curl 127.0.0.1:16726/order/1337/price/5
+curl 127.0.0.1:16726/order/1111/price/30
```
Vovke, line 1: thats not the way it reads, something like that instead:
`curl -X POST 'http://127.0.0.1:16726/v2/order/1111?currency=USDC&amount=30&callback=https://example.com/callback'`

src/server.rs
```diff
@@ -1,209 +1,258 @@
...
+                .ok_or_else(|| OrderError::MissingParameter(parameter.into()))
+        };
+
+        let currency = get_parameter(CURRENCY)?.to_owned();
+        let callback = get_parameter(CALLBACK)?.to_owned();
+        let amount = get_parameter(AMOUNT)?
+            .parse()
+            .map_err(|_| OrderError::InvalidParameter(AMOUNT.into()))?;
+
+        if currency != "USDC" {
```
Vovke, line 98: irrelevant

src/server.rs
```diff
@@ -1,209 +1,258 @@
...
+            return Err(OrderError::UnknownCurrency);
+        }
```
Vovke, line 100:
```suggestion

```

src/chain/tracker.rs
```diff
@@ -0,0 +1,303 @@
...
+
+                                            if invoice.death.0 >= now {
+                                                match invoice.check(&client, &watcher, &block).await {
+                                                    Ok(paid) => {
+                                                        if paid {
+                                                            state.order_paid(id.clone()).await;
+                                                        }
+
+                                                        id_remove_list.push(id.to_owned());
+                                                    }
```
Vovke, line 129:
```suggestion
                                                    }
                                                    Err(ChainError::StorageValueFormatNull) => {}
```

src/chain/tracker.rs
```diff
@@ -0,0 +1,303 @@
...
+                                                    Err(e) => {
+                                                        tracing::warn!("account fetch error: {0:?}", e);
```
Vovke, line 131: lots of noice about Paid being Null, need to ignore this error or preset Paid with value once invoice being created

Vovke:
lgtm

Vovke (approved):
lgtm

Issue #55, opened 2024-09-25
Finalize order response to be according to API specs


Vovke:
[API specs](https://alzymologist.github.io/kalatori-api/#/order/post_v2_order__orderId_)
```
{
  "order": "123456",
  "payment_status": "pending",
  "withdrawal_status": "waiting",
  "payment_account": "14E5nqKAp3oAJcmzgZhUD2RcptBeUBScxKHgJKU4HPNcKVf3",
  "amount": 1.5,
  "currency": {
    "currency": "DOT",
    "chain_name": "polkadot",
    "kind": "native",
    "decimals": 10,
    "rpc_url": "wss://rpc.polkadot.io"
  },
  "callback": "http://localhost:3000/api/kalatori/orders",
  "recipient": "15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6Sp5",
  "transactions": [
    {
      "block_number": 123456,
      "position_in_block": 1,
      "timestamp": "2021-01-01T00:00:00Z",
      "transaction_bytes": "0x1234567890abcdef",
      "sender": "14Gjs1TD93gnwEBfDMHoCgsuf1s2TVKUP6Z1qKmAZnZ8cW5q",
      "recipient": "15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6Sp5",
      "amount": 1.1,
      "currency": {
        "currency": "DOT",
        "chain_name": "polkadot",
        "kind": "native",
        "decimals": 10,
        "rpc_url": "wss://rpc.polkadot.io"
      },
      "status": "finalized"
    }
  ],
  "server_info": {
    "version": "0.2.0",
    "instance_id": "cunning-garbo",
    "debug": true,
    "kalatori_remark": "DEBUG=true;"
  }
}
```

Vovke:
we still need to have transactions on order so we could finish here. this part is postponed

Vovke:
blocked by #84

Vovke:
@fluiderson as you deal with #84, please ensure that this gets resolved as part of it. the idea is to have those transactions listed as part of the order response

PR #64, opened 2024-09-26
fix: order responses

Vovke:


Vovke:
resolves #67, #72 and part of #55 (still missing transactions, will be added in payment related PRs)

Slesarew (approved):

PR #75, opened 2024-10-08
fix: server_status returns instance_id

### LINKED ISSUE #73: Generate unique identifier for instance id and store it in database
The instance id should be generated when we run the daemon for the first time. 
instance_id should be stored in the database. 
We should fetch instance id from database if present, generate new one otherwise

Vovke:
resolves #73

### DIFF
--- a/src/database.rs
+++ b/src/database.rs
@@ -16,7 +16,7 @@ use crate::{
     error::{DbError, Error},
     task_tracker::TaskTracker,
 };
-use names::Generator;
+use names::{Generator, Name};
 use parity_scale_codec::{Decode, Encode};
 use std::time::SystemTime;
 use substrate_crypto_light::common::AccountId32;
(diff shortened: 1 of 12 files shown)


src/database.rs
```diff
@@ -230,6 +235,40 @@ impl Database {
...
+                                let server_info: ServerInfo =
+                                    serde_json::from_slice(&server_info_data).map_err(|e| {
+                                        DbError::DeserializationError(e.to_string())
+                                    })?;
+                                Ok(server_info.instance_id)
+                            } else {
+                                let mut generator = Generator::default();
+                                let new_instance_id = generator
+                                    .next()
+                                    .unwrap_or_else(|| "unknown-instance".to_string());
```
fluiderson, line 255:
```suggestion
                                    .unwrap();
```
The default generator always returns `Some()`.

fluiderson (dismissed):

src/database.rs
Slesarew: This could be changed to meaning error in case namegen decides to use some awkward unreliable rng backend in future and we forget it, but this fallback seems reasonable to me at least for now. Or at least expect-style panic would work to explain why we just unwrap here.

src/database.rs
```diff
@@ -245,6 +284,12 @@ impl Database {
         Ok(Self { tx })
     }
 
+    pub async fn initialize_server_info(&self) -> Result<String, DbError> {
```
Slesarew, line 287: why not just do it at db init? We do it every time, there is not much sense to do this in parallel, we do it only once, if this fails we should not be running this system anyway.

Slesarew (approved):

Slesarew (changes requested):
wait, bump version!

Vovke:
Resolves #76

Vovke:
Resolves #75

Vovke:
Resoves #59

Slesarew (approved):

PR #80, opened 2024-10-10
docs: database structure

Vovke:
Want to discuss database structure here


docs/DATABASE.md
```diff
@@ -0,0 +1,32 @@
+Plan is to update the database scheme in a way that it will support the requirements we have as for the API specs and additional improvements of the deamon.
+
+## Tables
+### Orders (`orders`)
+- order - String: order identifier provided by the frontend 
+- payment_status - Enum: (pending|paid|timed_out). 
+- withdrawal_status - Enum: (waiting|failed|completed|none).
```
Slesarew, line 7: we might need "forced" status as well

docs/DATABASE.md
```diff
@@ -0,0 +1,32 @@
+Plan is to update the database scheme in a way that it will support the requirements we have as for the API specs and additional improvements of the deamon.
```
Slesarew, line 1: are these things derived from API? Please indicate if something is different from types defined there, ideally these should be as similar as possible, but deviations are ok where they make sense

docs/DATABASE.md
```diff
@@ -0,0 +1,32 @@
...
+- callback: String: Callback url for frontend order status update 
+- payment_account: Derived address for this order. 
+- recipient: Address that will receive the payout once the order is fulfilled. 
+- death: Expiry timestamp for the order.
+
+### Transactions (`transactions`)
+- order - String: order id to link transaction to order
+- chain - String: identifier for the chain where transaction occured
+- block_nmber - Integer: Block number where the transaction is recorded.
+- position_in_block - Integer: Position of the transaction within the block.
```
Slesarew, line 19: we might also (later?) want to store the block for which transaction was submitted. It is included into saved transaction bytes though, so maybe not

docs/DATABASE.md
Vovke: Addressed in [c69aa99](https://github.com/Alzymologist/Kalatori-backend/pull/80/commits/c69aa99b6747242f1a30f2348185c35d147486d8)

docs/DATABASE.md
```diff
@@ -0,0 +1,35 @@
+Plan is to update the database scheme in a way that it will support the requirements we have as for the API specs and additional improvements of the deamon.
+
+## Tables
+### Orders (`orders`)
+- order - String: order identifier provided by the frontend 
+- payment_status - Enum: (pending|paid|timed_out). 
+- withdrawal_status - Enum: (waiting|failed|completed|forced|none). 
+- amount - Float: Order amount
```
Vovke, line 8:
```suggestion
- amount - u128: Order amount 
```

docs/DATABASE.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Tables
+### Orders (`orders`)
+- order - String: order identifier provided by the frontend 
+- payment_status - Enum: (pending|paid|timed_out). 
+- withdrawal_status - Enum: (waiting|failed|completed|forced|none). 
+- amount - Float: Order amount 
+- currency - String: Currency ticker ("DOT"|"USDC"|...). 
+- callback: String: Callback url for frontend order status update 
+- payment_account: String: Derived address for this order.
```
Vovke, line 11:
```suggestion
- payment_account: [u8; 32]: Derived address for this order. 
```

docs/DATABASE.md
```diff
@@ -0,0 +1,35 @@
...
+## Tables
+### Orders (`orders`)
+- order - String: order identifier provided by the frontend 
+- payment_status - Enum: (pending|paid|timed_out). 
+- withdrawal_status - Enum: (waiting|failed|completed|forced|none). 
+- amount - Float: Order amount 
+- currency - String: Currency ticker ("DOT"|"USDC"|...). 
+- callback: String: Callback url for frontend order status update 
+- payment_account: String: Derived address for this order. 
+- recipient: String: Address that will receive the payout once the order is fulfilled.
```
Vovke, line 12:
```suggestion
- recipient: [u8; 32]: Address that will receive the payout once the order is fulfilled. 
```

docs/DATABASE.md
```diff
@@ -0,0 +1,35 @@
...
+- withdrawal_status - Enum: (waiting|failed|completed|forced|none). 
+- amount - Float: Order amount 
+- currency - String: Currency ticker ("DOT"|"USDC"|...). 
+- callback: String: Callback url for frontend order status update 
+- payment_account: String: Derived address for this order. 
+- recipient: String: Address that will receive the payout once the order is fulfilled. 
+- message: String|null: Optional parameter for failed orders.
+- payment_page: String|null: Optional parameter for the frontend to redirect to a payment page.
+- redirect_url: String|null: Optional parameter for the frontend to redirect once the order is repaid.
+- death: Timestamp: Expiry timestamp for the order.
```
Vovke, line 16:
```suggestion
- death: u32: Expiry timestamp for the order.
```

Vovke:
@fluiderson thanks a lot for wrapping it up, I've addressed the remarks rather below or with code suggestions.

>Firstly, tables aren't enough. We also need to define keys for them.

the keys are 
- orders table will be order_id
- transactions table, transaction_id
- instance_info table, instance_id

> Now about the scheme itself. It'd be nice to have descriptions for enum variants, too.

My goal with this Pull Request is to have a discussion and decide as for database structure, we can describe here enums as well, but thats not the goal atm.


# `orders`

> - `currency`
Why not make it as it's in the API specification? On Substrate blockchains, currency isn't defined as a string. There's always a single native token, and sometimes they have multiple assets that're defined by an ID number (`u32`, precisely). They have a name, but it's just metadata that can be changed anytime. So, a name is a bad value for a key of currency.

thats the way we get it from frontend, even assuming the small probability of ticker name update, we will still get same ticker from there. I agree that this is not the best idea to store it that way, but thats something we will take care of in the future. Created https://github.com/Alzymologist/Kalatori-backend/issues/89 to take care of it later.

> - `payment_page` & `redirect_url`
It isn't clear from where the daemon should receive these fields, and hence whether it should store them at all.

those fields are listed in the API, some integrations expected to require those fields

Also, we may need to link an order to the blockchain that it's on. So, I propose adding the `chain` field with the type `H64`.

# `transactions`

> The key for this table can't be just an order ID. I propose almost the same scheme as Substrate blockchains are using for transactions. The key should consist of 3 parts: the order ID, the block hash, & the transaction ID in the block. But there's a hitch, we can't know in which block the transaction will end up & what ID it'll have until the block is recorded on a blockchain.  

I suggest to use transaction id as a key, and adding array of transactions to order table or introducing "order_id to transaction array table", wdyt?

> So, what the `pending` status is for? I propose to get rid of it because we already have `payment_status` & `withdrawal_status` in the `orders` table. If our transfer transaction will go wrong and won't be recorded on the blockchain, we can record it in the `message` field, or simply print it in the log. Or both.

pending status is initial transaction status for transactions which are yet to be included in the block. That is listed in the API specs as well.

> - `chain`
Remove this field. An order has it.
>- `block_number`
Use block hash in the key instead of the block number.
>- `position_in_block`
A part of the key.
>- `transaction_bytes`
Remove this field. A transaction can be fetched from a blockchain.
>- `timestamp`
Remove this field. A timestamp is included in the transaction that can be fetched from a blockchain.
>- `sender`.
Remove this field. See above.
>- `recipient`.
Remove this field. See above.
>- `amount`
Remove this field. See above.
>- `currency`
Remove this field. See above.

While this information is indeed available on chain I would prefer to store it in the db for 3 reasons:
- in case we want to provide some kind of statistics, csv report or any action which require fetching all the transaction data from the database it will make it way harder to populate it
- as we will eventually migrate to a more convenient data storage having those fields will make it easier for us to migrate
- its way easier to implement it that way while also minimizing the amount of race condition presented by rpc instability e.g I should always have enough information about transactions especially in cases where rpcs are unstable and doesn't allow me to connect

# `instance_info`

The only field I need there is instance name, all the other fields are stored without specific reason.

# `orders_per_chain`
## K: `H64` V: `u64`

Please create an issue to add this table if its needed

docs/DATABASE.md
```diff
@@ -0,0 +1,35 @@
+Plan is to update the database scheme in a way that it will support the requirements we have as for the API specs and additional improvements of the deamon.
+
+## Tables
+### Orders (`orders`)
+- order - String: order identifier provided by the frontend
```
Vovke, line 5:
```suggestion
- order - String: order identifier provided by the frontend 
- chain - H64: chain name where order is planned to be repaid
```

fluiderson:
> I suggest to use transaction id as a key

What's the transaction ID? Do you mean the extrinsic ID from Subscan?

> and adding array of transactions to order table or introducing "order_id to transaction array table", wdyt?

We can store transactions in a single table as per my key design. What're the benefits of adding another table or field in the `orders` table?

> pending status is initial transaction status for transactions which are yet to be included in the block. That is listed in the API specs as well.

To store a transaction, we need its ID. Until it isn't included in the block, what ID should we use for it, then? Either this is a fundamental flaw in our API spec or we need to come up with a more complex key to identify both included & pending transactions.

> The only field needed there is instance name, all the other fields are stored without specific reason.

So, the current key checking, the key rotation mechanism, the chains genesis checking will be in future versions, too? If yes, then please remove the `chain` field from the `orders` table, because it's a part of the chains genesis checking. `H64` can't be generated without it.

Vovke:
@fluiderson I definitely would prefer having extrinsic but as you said that won't allow us to precreate the transaction.
As discussed above we can have an array of linked transaction id inside the order table or having linking table order_id -> [trx_id_1, trx_id_2, trx_id_3....], both of those approaches will allow us to have our own transaction id as a key for fetching correct transactions for the order. 

> If yes, then please remove the chain field from the orders table, because it's a part of the chains genesis checking. H64 can't be generated without it.

removed

Vovke:
discussed with @fluiderson 
- we will generate our own `transaction_id` to link transaction and order
- need to add transaction `type` field to the transactions, can be `payment|withdrawal`

docs/DATABASE.md
```diff
@@ -0,0 +1,35 @@
...
+- callback: String: Callback url for frontend order status update 
+- payment_account: [u8; 32]: Derived address for this order. 
+- recipient: [u8; 32]: Address that will receive the payout once the order is fulfilled. 
+- message: String|null: Optional parameter for failed orders.
+- payment_page: String|null: Optional parameter for the frontend to redirect to a payment page.
+- redirect_url: String|null: Optional parameter for the frontend to redirect once the order is repaid.
+- death: u32: Expiry timestamp for the order.
+
+### Transactions (`transactions`)
+- order - String: order id to link transaction to order
```
Vovke, line 19:
```suggestion
- transaction_id - unique id generated by us to allow linking transaction to order
- order - String: order id to link transaction to order
```

docs/DATABASE.md
```diff
@@ -0,0 +1,35 @@
...
+- chain - String: identifier for the chain where transaction occurred
+- block_number - Integer: Block number where the transaction is recorded.
+- position_in_block - Integer: Position of the transaction within the block. 
+- timestamp - Timestamp: Timestamp of the transaction. 
+- transaction_bytes - String: Raw transaction data. 
+- sender - String: Address sending the transaction. 
+- recipient - String: Address receiving the transaction. 
+- amount - Float: transaction amount
+- currency: String: Transaction currency 
+- status - Enum: Transaction status (pending|finalized|failed).
```
Vovke, line 29:
```suggestion
- type - Enum: Transaction type (payment|withdrawal) to distinguish between internal (withdrawal) and external (payment) transactions
- status - Enum: Transaction status (pending|finalized|failed).
```

Vovke:
@Slesarew this can be merged

Slesarew (approved):

PR #97, opened 2024-10-23
feat: simple callback

### LINKED ISSUE #63: Implement callback functionality

Vovke:
Resolves #63


src/state.rs
```diff
@@ -132,7 +132,17 @@ impl State {
                                 // Only perform actions if the record is saved in ledger
                                 match state.db.mark_paid(id.clone()).await {
                                     Ok(order) => {
-                                        // TODO: callback here
+                                        if let Some(callback) = order.callback.clone() {
+                                            tokio::spawn(async move {
+                                                tracing::info!("Sending callback to: {}", callback);
+                                                let client = reqwest::Client::new();
```
Slesarew, line 138: is this like 3rd different entity called just "client" here?

Vovke: addressed in 4576c71

Cargo.toml
```diff
@@ -63,6 +63,7 @@ substrate_parser = { git = "https://github.com/Alzymologist/substrate-parser" }
 substrate-constructor = { git = "https://github.com/Alzymologist/substrate-constructor" }
 mnemonic-external = { git = "https://github.com/Alzymologist/mnemonic-external" }
 substrate-crypto-light = { git = "https://github.com/Alzymologist/substrate-crypto-light" }
+reqwest = "0.12.8"
```
fluiderson, line 66:
```suggestion
reqwest = "0.12"
```
Let's specify only the latest major version. It does the same & looks conciser.

fluiderson (commented):
https://github.com/Alzymologist/Kalatori-backend/blob/56cda51716ee0d11491b031838503380047f6fa3/Cargo.toml#L23
Remove this HTTP client crate from deps if you decided to use `reqwest` instead.

Cargo.toml
Vovke: addressed in 4cc2713

fluiderson (approved):

Vovke:
included in #106

PR #105, opened 2024-10-29
fix: unbreak multisig

### LINKED ISSUE #78: Payout with DOT failing on both prod and test chains

Slesarew:
eventually should resolve #78 and #79


Slesarew:
#106 included this

PR #110, opened 2024-10-31
feat: order force withdrawal

Vovke:


fluiderson (dismissed):
All fine, but do we need this in the main branch now if this just marks an order as withdrawn without actually withdrawing from it?

Vovke:
> All fine, but do we need this in the main branch now if this just marks an order as withdrawn without actually withdrawing from it?

@fluiderson that part still WIP, way more work than expected

src/chain/payout.rs
```diff
@@ -74,40 +77,35 @@ pub async fn payout(
...
                     )?]
                 }
-            },
-            a if (loss_tolerance..=manual_intervention_amount).contains(&a) => {
-                tracing::warn!("Overpayment, proceeding with available balance");
-                // We will transfer all the available balance
-                // TODO smarter handling and returns probably
+            }
+        } else {
+            tracing::warn!("Overpayment or forced");
```
Vovke, line 82: change to info

Vovke, line 82:
```suggestion
            tracing::info!("Overpayment or forced");
```

src/chain/payout.rs
```diff
@@ -43,15 +43,18 @@ pub async fn payout(
         let block_number = current_block_number(&client, &chain.metadata, &block).await?;
         let balance = order.balance(&client, &chain, &block).await?; // TODO same
         let loss_tolerance = 10000; // TODO: replace with multiple of existential
-        let manual_intervention_amount = 1000000000000;
+                                    // let manual_intervention_amount = 1000000000000;
```
Vovke, line 46:
```suggestion
        // let manual_intervention_amount = 1000000000000;
```

Slesarew:
```suggestion
                                    // TODO: add upper limit for transactions that would require manual intervention
                                    // just because it was found to be needed with non-crypto trade, who knows why?
```

Vovke:
linking issue for upper tx limit
https://github.com/Alzymologist/Kalatori-backend/issues/124

src/chain/tracker.rs
```diff
@@ -243,22 +253,39 @@ impl ChainWatcher {
...
-            tracing::info!(
-                "chain {} has token {} with properties {:?}",
-                &chain.name,
-                &name,
-                &properties
-            );
-            if let Some(native_token) = &chain.native_token {
-                (native_token.name == *name) && (native_token.decimals == specs.decimals)
-            } else {
+        assets = assets
```
Slesarew, line 256: why replace smart `retain` and `any` with primitives?

Vovke: @fluiderson ^^

fluiderson: @Slesarew, this code takes names for assets from the config, not from a chain. The `retain` method doesn't allow the mutating of the key value, hence `filter_map` and then `collect`.

Slesarew: and why not from the chain? Taking those from chain ensures they will be accepted by chain and things will match as advertised, does it not?

Slesarew (approved):

src/chain/tracker.rs
fluiderson: > and why not from the chain? Taking those from chain ensures they will be accepted by chain and things will match as advertised, does it not?
Basically, I just removed checks for asset names from your code. Everything else is the same.

fluiderson: If you don't get why, there's no such thing as an asset name for the Assets pallet in terms of the asset transferring, just IDs. Names are just a metadata for humans there. Now, the daemon user can set the same asset names as on the chain or name them on their own (with the checks, this was impossible). The latter feature was very much requested for testing purposes by the frontend team.

PR #135, opened 2024-11-15
fix: remove routine events check

Slesarew:
There used to be a significant optimisation that watched events and checked only accounts that have tranfers into them in given block; this was found to be unreliable: there are ways to transfer funds without emitting a transfer event (one notable example is through asset exchange procedure directed straight into invoice account), and probably even without any reliably expected event (through XCM). Thus we just scan all accounts, every time. Please submit a PR or an issue if you figure out a reliable optimization for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.7] - 2024-11-18
+
+### 🚀 Features
+
+- Asset Hub transactions with fee currency
+  - Autofill tip with asset
+  - Pass asset id into transaction constructor to properly select fee currency
+
+### 🧪 Testing
+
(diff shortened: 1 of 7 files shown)


Vovke:
added more test cases and fixed incorrect invoice check to trigger correctly

@Slesarew there are some issue with lint in tracker.rs as you run cargo fmt. please fix it, bump version and lets merge it. per functionality it works correctly for the cases provided.

Vovke (approved):
added more test cases and fixed incorrect invoice check to trigger correctly

@Slesarew there are some issue with lint in tracker.rs as you run cargo fmt. please fix it, bump version and lets merge it. per functionality it works correctly for the cases provided.

Slesarew:
Aww I can't believe it, XCM format changed again. Turns out these are really good tests!

Vovke (approved):

PR #197, opened 2025-09-14
feat: metadata v16 support

Vovke:


kirushik:
Force-approving this as a quickfix for the existing live deployments, @Vovke please fix all the lints and reports as your next PR

PR #215, opened 2025-10-02
refactor: migrate chain watcher to subxt

AnlisUA:
This PR was branched out from [refactor: use only asset-hub (statemint) chain](https://github.com/Kalapaja/Kalatori/pull/214) and contains it's changes, please review this one only after merge of previous PR

### DIFF
--- a/src/chain/payout.rs
+++ b/src/chain/payout.rs
@@ -6,180 +6,135 @@
 
 use crate::{
     chain::{
-        definitions::Invoice,
-        rpc::{block_hash, current_block_number, send_stuff},
-        tracker::ChainWatcher,
-        utils::{
-            construct_batch_transaction, construct_single_asset_transfer_call,
-            construct_single_balance_transfer_call, AssetTransferConstructor,
-            BalanceTransferConstructor,
-        },
+        definitions::Invoice, runtime::runtime_types::{staging_xcm::v3::multilocation::MultiLocation, xcm::v3::{junction::Junction, junctions::Junctions}}, tracker::ChainWatcher, AssetHubConfig, AssetHubOnlineClient
     },
     database::{TransactionInfoDb, TransactionInfoDbInner, TxKind},
     definitions::{
-        api_v2::{Amount, TokenKind, TxStatus},
-        Balance,
+        api_v2::{Amount, TxStatus},
     },
     error::ChainError,
-    signer::Signer,
     state::State,
 };
-use frame_metadata::v15::RuntimeMetadataV15;
-use jsonrpsee::ws_client::WsClientBuilder;
-use substrate_constructor::fill_prepare::TypeContentToFill;
+use base58::ToBase58;
 use substrate_crypto_light::common::AsBase58;
+use subxt::config::DefaultExtrinsicParamsBuilder;
+use subxt_signer::{bip39::Mnemonic, sr25519::Keypair, DeriveJunction, ExposeSecret, SecretString};
+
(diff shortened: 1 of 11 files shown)


src/chain/payout.rs
```diff
@@ -6,180 +6,135 @@
...
+        interior: Junctions::X2(Junction::PalletInstance(50), Junction::GeneralIndex(asset_id as u128)),
+    };
+
+    let tx_config = DefaultExtrinsicParamsBuilder::<AssetHubConfig>::new()
+        .tip_of(0, location)
+        .mortal(32)
+        .build();
+
+    // TODO: need to validate phrase on start so unwrap here can be safe
+    let mnemonic = Mnemonic::parse(seed.expose_secret()).unwrap();
```
Vovke, line 98: mb new error types like ChainError::InvalidMnemonic and ChainError::KeypairGeneration?

src/chain/tracker.rs
```diff
@@ -312,87 +402,88 @@ pub fn start_chain_watch(
...
+
+            let Some(response) = client
+                .storage()
+                .at_latest()
+                .await?
+                .fetch(&request_data)
+                .await?
+                else {
+                    // TODO: panic or work without this asset? Need to notify user about error somehow
+                    panic!("Asset {} with id {} not found on chain {}", asset.name, asset.id, chain.name)
```
Vovke, line 460: I'd return Error here instead of panic, the user will eventually notice it from the frontend side once the asset won't come back. Although I don't see another reason for this but misconfiguration

src/chain/tracker.rs
```diff
@@ -137,34 +202,44 @@ pub fn start_chain_watch(
...
-                                    &watcher.metadata,
-                                )
-                                    .await {
-                                        Ok((timestamp, events)) => {
-                                        tracing::debug!("Got a block with timestamp {timestamp:?} & events: {events:?}");
+                                ).await {
+                                    Ok((timestamp, extrinsics)) => {
+                                        tracing::debug!("Got a block with timestamp {timestamp:?} & {} extrinsics", extrinsics.len());
+
+                                        // TODO: handle Err results? Log them at least?
```
Vovke, line 211: I think we can do one iteration here instead directly on chain.
also agree on not silently ignoring the errors

src/chain/payout.rs
```diff
@@ -6,180 +6,135 @@
...
+    // TODO: add support for password configuration, ensure unwrap here is safe
+    let keypair = Keypair::from_phrase(&mnemonic, None).unwrap();
+
+    let order_keypair = keypair.derive([
+        DeriveJunction::hard(order.recipient.to_base58_string(2)),
+        DeriveJunction::hard(&order.id),
+    ]);
+
+    // TODO: why 42? perhaps store it into constant?
```
Vovke, line 107: 👍🏾

Vovke (approved):
some minor issues, looks good overall

src/chain/payout.rs
AnlisUA: yep, added new errors (partially replaced old ones) in https://github.com/Kalapaja/Kalatori/pull/219

src/chain/tracker.rs
AnlisUA: Guess we'll need to discuss in which cases we should panic and in which we shouldn't. Yeah, it's definitely misconfiguration case, probably we'd better check it out somewhere on the start

PR #216, opened 2025-10-05
refactor: configs

AnlisUA:
This PR was branched out from https://github.com/Kalapaja/Kalatori/pull/215 and contains it's changes, please review this one only after merge of previous PR


src/configs.rs
```diff
@@ -0,0 +1,350 @@
...
+pub struct WebServerConfig {
+    /// By default use 0.0.0.0
+    #[serde(default = "default_host")]
+    pub host: IpAddr,
+    /// By default use port 16726
+    #[serde(default = "default_port")]
+    pub port: u16,
+}
+
+fn dafault_database_path() -> String {
```
Vovke: typo

src/main.rs
```diff
@@ -76,63 +77,48 @@ fn main() -> ExitCode {
...
 
-        None
-    } else {
-        Some(config.database.unwrap_or_else(|| DATABASE_DEFAULT.into()))
-    };
+    let seed_config = seed_config_with_prefix(&configs_path, &env_prefix);
+
+    let chain_config = chain_config_with_prefix(&configs_path, &env_prefix);
+    let payments_config = payments_config_with_prefix(&configs_path, &env_prefix);
+    let web_server_config = web_server_config_with_prefix(&env_prefix, &configs_path);
```
Vovke: sudden parameter order change

Vovke:
the seed is now stored in plain text JSON file, security regression from prev CLI based approach

Vovke (approved):
good overall, methioned security issue regards seed being stored in plain text

src/main.rs
AnlisUA: yep, it was a bug, fixed

src/configs.rs
AnlisUA: fixed

AnlisUA:
> the seed is now stored in plain text JSON file, security regression from prev CLI based approach

Actually, it's not. Taking into account that we expect running the application in docker container, this CLI parameter in any way should be passed there over env var. Updated approach also supports env vars, it's still valid option to pass `KALATORI_SEED_SEED` env var into container, application will handle it. It's not the recommended way to pass secrets into containerized environments though. In general a bit better way is to mount secret as a volume, see [link](https://cheatsheetseries.owasp.org/cheatsheets/Secrets_Management_Cheat_Sheet.html#51-injection-of-secrets-file-in-memory) for details.

PR #246, opened 2026-02-10
feat: add URL validation with SSRF prevention for redirect/image URLs

### LINKED ISSUE #238: Validate URLs in API
# Description
We need to validate URLs which we receive in requests. At this moments there are 2 types of URLs we want to validate:
1. Redirect url, used in `CreateInvoiceParams`
2. Image url, used in `InvoiceCartItem`

## Validations
We want to validate that:
- URL is actually valid (has valid schema)
- Doesn't contain suspicious patterns (like `@` or `\\` symbols)
- Contains allowed domain
- (For images) ends with `.png`, `.jpg`, `.jpeg` etc (please search for popular image extensions)
- Perhaps there are other good practices, please google it or ask LLMs

## About domain validation
Alongside with URLs validations, we need to extend `ShopConfig` with new optional parameters: `allowed_redirect_url` (basic url, can be just domain) and `allowed_image_urls` (list of basic urls/domains in case if the shop uses CDN). Both parameters are optional and by default are set to domain part of `invoices_webhook_url` parameter of the same config.

techraed:
resolves #238 

The PR implements URL validation for redirect/image URLs provided with create_invoice and update_invoice (only for image URL as a part of cart object).

Key points:
1. `ShopConfig` now has a customized deserialization which actually checks that provided allowed redirect/image base URL are real proper URLs, not just random `String`s. Checks include port validation, protocol (https) validation, domain existence as a host part of the URL.
2. Url validation functions are `validate`, `validate_with_allowed_base` and `validate_with_allowed_base_many`. The latter perform base url checks along with a common validation routine.
3. Url full validation routine:
    - **Length restriction** - Maximum 2048 characters
    - **URL parsing** - Must conform to WHATWG URL Standard
    - **HTTPS-only** - Rejects `http://`, `file://`, `ftp://`, etc.
    - **Port restriction** - Only allows port 443 (HTTPS default)
    - **No credentials** - Rejects `user:password@host` in URL
    - **Base URL restriction** - Ensures the URL matches (is the same or a subdomain of) the allowed base URL if provided (domain validation)
    - **DNS resolution + IP validation (ipv4/ipv6)** - Performs real-time DNS lookup and validates that resolved IP addresses are:
        - Not loopback (`127.0.0.0/8`, `::1`)
        - Not private (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`)
        - Not link-local (`169.254.0.0/16`) - blocks cloud metadata endpoints
        - Not CGNAT (`100.64.0.0/10`) - blocks Alibaba Cloud metadata
        - Not documentation/test ranges (`192.0.2.0/24`, `198.51.100.0/24`, etc.)
        - Not multicast, broadcast or reserved ranges
4. Database stores only validated urls. `ValidatedUrl` can store in the DB as it implements necessary `sqlx` traits. However, `Decode` implementation doesn't re-validate the URL as we assume that all URLs in the database have been validated before insertion.

TODO:
- [x] add domain validation
- ~~[ ] check locally on e2e tests~~


AnlisUA:
Hi @techraed and thanks for your contribution!

Unfortunately, I completely forgot to mention that our main branch currently contains the legacy version of the Kalatori daemon, while we're actively developing the new one under the develop branch. We haven't released the new version yet, so we haven't merged it into main for that reason. The new version is quite significantly different from what you can see in main now.

From what I see, your PR mostly concentrates on brand new URL validation logic, and we'll be able to keep it, but some of the integration/calling code will need to be reworked for the new version.

techraed:
hey @AnlisUA 
Thank you for a quick response

Could you guide me what's the main working branch right now, so I update the code?

AnlisUA:
> hey @AnlisUA Thank you for a quick response
> 
> Could you guide me what's the main working branch right now, so I update the code?


Sure, we're working in [develop](https://github.com/Kalapaja/Kalatori/tree/develop) branch

daemon/src/utils/url_validation.rs
```diff
@@ -0,0 +1,450 @@
...
+            "https://example.com:443/path",
+            "https://example.com/path?order=123&status=paid",
+            "https://8.8.8.8/webhook",
+            "https://1.1.1.1/webhook",
+            "https://93.184.216.34/webhook",
+        ];
+        for url in urls {
+            assert!(validate_and_resolve(url).await.is_ok());
+        }
+    }
```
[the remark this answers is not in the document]
techraed: 1. we expect CI being able to perform net requests.
2. example.com is a reserved name in RFC 2606, it should be able to resolve it

src/handlers/order.rs
```diff
@@ -47,10 +50,20 @@ pub async fn process_order(
...
 
+        // CALLBACK validation
+        let mut callback = payload.callback.unwrap_or_default();
+        if !callback.is_empty() {
+            let ip = url_validation::validate_and_resolve(&callback)
+                .await
+                .map_err(OrderError::InvalidCallback)?;
+
+            callback = ip.to_string();
+        }
```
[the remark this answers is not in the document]
techraed: done after rebase

daemon/src/utils/url_validation.rs
```diff
@@ -0,0 +1,805 @@
...
+where
+    DB: Database,
+    String: Decode<'r, DB>,
+{
+    fn decode(value: <DB as Database>::ValueRef<'r>) -> Result<Self, BoxDynError> {
+        let url = String::decode(value)?;
+        let parsed = Url::parse(&url)?;
+        Ok(Self(parsed))
+    }
+}
```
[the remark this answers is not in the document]
techraed: this is a known/mentioned compromise
`ValidatedUrl` is only generated after url validation and then stored/retrieved from the DB
There's no way to store invalid url in the DB.

daemon/src/utils/url_validation.rs
```diff
@@ -0,0 +1,805 @@
...
+
+        if let Err(reason) = check_ipv4_is_global(ipv4) {
+            return Err(UrlValidationError::NonGlobalIp {
+                hostname: host.to_string(),
+                ip: ip.to_string(),
+                reason,
+            });
+        }
+
+        maybe_ip.get_or_insert(ip);
```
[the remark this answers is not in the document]
techraed: added ipv6 checks

daemon/src/state.rs
```diff
@@ -523,6 +446,208 @@ impl<D: DaoInterface> AppState<D> {
...
+                    image_url,
+                    &self.allowed_image_domains,
+                )
+                .await
+                .map_err(
+                    |source| DaoInvoiceError::InvalidUrlParameter {
+                        description: "Invalid cart item image URL",
+                        source,
+                    },
+                )?;
```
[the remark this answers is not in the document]
techraed: the `.png` (and other) does not guarantee the response is an image.
wdyt @AnlisUA ?

daemon/src/state.rs
```diff
@@ -523,6 +446,208 @@ impl<D: DaoInterface> AppState<D> {
...
+                quantity: item.quantity,
+                price: item.price,
+                tax: item.tax,
+                discount: item.discount,
+                image_url,
+                product_url,
+            };
+
+            ret_items.push(cart_item);
+        }
```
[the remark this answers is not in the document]
techraed: done

techraed:
@AnlisUA 
I guess it's ready for a review now:
- rebased on `develop`
- fully resolved an issue with two things that are good to be clarified:
    - this https://github.com/Kalapaja/Kalatori/pull/246#discussion_r2819761223
    -is there domain validation (allowed domains) for products urls?

daemon/src/state.rs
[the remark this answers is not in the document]
AnlisUA: Yeah, true. But as long as we receive this data from the shop, this validation is more about notifying the shop if they put here something wrong/unexpected.

daemon/src/configs/types.rs
```diff
@@ -329,15 +334,147 @@ pub struct ShopMetaConfig {
...
 
-#[derive(Deserialize, Clone)]
+#[derive(Clone)]
 pub struct ShopConfig {
-    pub invoices_webhook_url: String,
-    #[serde(default = "default_signature_max_age_secs")]
+    pub invoices_webhook_url: Url,
     pub signature_max_age_secs: u64,
-    #[serde(flatten)]
+    pub allowed_base_redirect_url: Option<Host<String>>,
```
AnlisUA: I'm afraid it might be not obvious why we have one field name in the structure and the other in the actual config. Perhaps it's gonna be easier to leave this fields "raw" but add some method which will refine it all into another structure, something like:
```
struct UrlValidatorConfig {
    allowed_redirect_base_url: Host<String>,
    allowed_base_image_urls: Vec<Host<String>>,
}
```

Another thing which will be nice to have in the config is something like `allow_unsecure_urls` with default `false` value. It might be useful for manual testing in local environment with some e-commerce platforms we'll develop integration for.

daemon/src/configs/types.rs
```diff
@@ -329,15 +334,147 @@ pub struct ShopMetaConfig {
...
+            allowed_image_urls: Option<Vec<Url>>,
+            #[serde(flatten)]
+            meta: ShopMetaConfig,
+        }
+
+        let raw = RawShopConfig::deserialize(deserializer)?;
+
+        // Validate url is of https protocol, port 443 (if has any) and has a domain
+        // host.
+        fn validate_https_443(url: &Url) -> Result<(), String> {
```
AnlisUA: I suppose that we can also move this checks into the method which will convert "raw" values into the new structure I described in another comment. TBH I'm not sure that custom deserializer is a good idea, it's harder to test and sometimes might make some thing not obvious (like different field names).

Another thing I found is that some of this checks are pretty similar to the ones you have in `utils/url_validation.rs` in methods like `validate_with_allowed_base_impl`, perhaps it makes sense to have some bunch of low level checker-functions which then can be reused in both config validation and runtime URLs validations?

daemon/src/dao/invoice.rs
```diff
@@ -61,6 +62,36 @@ pub enum DaoInvoiceError {
     /// Database operation failed
     #[error("Database error during invoice operation")]
     DatabaseError,
+
+    #[error("Invalid URL parameter: {description}, error: {source:?}")]
+    InvalidUrlParameter {
```
AnlisUA: I believe that we don't need such kind of validations (both `InvalidUrlParameter` and `DnsLookupFailed`) for data loaded from database. We do not and do not plan to provide any options to store any data into database except of over API. WIth that said I believe that we can only validate the URLs on input (in API handlers) and even keep them in structures as raw strings. You're right that some url can become inaccessible over time and DNS lookup might fail, but on the other hand we'll definitely want to be able to fetch those records from admin panel and see them without errors even if the domain of the shop has been changed or something else has happened.

UPD: found your comment that we don't validate it on load from database, so it's another reason to get rid of this error kinds.

daemon/src/types/invoice.rs
```diff
@@ -39,7 +44,7 @@ pub struct Invoice {
     pub payment_address: String,
     pub status: InvoiceStatus,
     pub cart: InvoiceCart,
-    pub redirect_url: String,
+    pub redirect_url: ValidatedUrl,
```
AnlisUA: As I said in another comment, I don't see big reasons to change field types here. It forces us to add a lot of similar boilerplate code but the only thing it gives is assurance that this url has been validated. On the other hand it doesn't deny any other validations which we'll have to add later (for instance, validate that provided `Decimal` price is positive at least).

My idea is that we can "believe" that all data which circles inside the daemon is valid by-default. The data which comes from 3d-party services (API) might be invalid and should be validated on API handlers level.

AnlisUA (changes requested):
Overall, the great work has been done but I'm not sure about some decisions. Probably I had to describe my vision in the issue, sorry for misunderstanding.

The story is that I see a bit different data validation approach with such principles:
1. By default we expect that in configs/API/somewhere else we can receive invalid/incorrect data and have to validate it
2. When the received data has been validated, we don't need to validate it again on any step. Important clarification: we treat data loaded from database as "internal", only daemon can store something in database (without manual intervention or other similar kind of staff we're not responsible for), so as long as we validate the data on "inputs" before storing it to database, the loaded data is always valid
3. Data inputs should validate it. In our case data inputs are configs (some of which are additionally validated on startup, check `main.rs` calls) and API.
4. I think that we should keep data validation on this "inputs" (or at least close to them). For example, we validate configs by the `serde` itself + on program startup we're calling some separate methods to validate/refine them. I'd like to have something similar with API. I see it in the next way:
- We can update `ApiState` type from type alias to structure which will store `AppState` with another field, something like `validator: ApiParamsValidator`
- In api handler functions we can call methods of this new `ApiParamsValidator` structure in order to validate incoming params. Let's start from URLs validations, later we'll add the others
- We can keep this `ApiParamsValidator` somewhere under `api` module, no need to expose it somewhere else
- On program startup we can call some function on the shop config which will return us some refined structure which will already contain only required fields for further validation. Then this structure can be passed into `api_server` initializer function
- In that way we'll also keep `State` clean from validation logic which might be useful in future, for instance, when we'll add some `dev` handlers (for debugging purposes) which will skip validations in most cases

daemon/src/configs/types.rs
techraed: > I suppose that we can also move this checks into the method which will convert "raw" values into the new structure I described in another comment.

Although I truly believe having a custom deserialization is good thing here, because it gives us natively required guarantees, I totally agree with you here - custom deserialization is odd. The validation approach you described in other comments makes configs/API as sources of some abstract `DataInput`, that can implement validation like `impl Validate for DataInput` in the future.

Will do as you requested.

daemon/src/configs/types.rs
techraed: done

daemon/src/configs/types.rs
techraed: done

daemon/src/dao/invoice.rs
techraed: done

daemon/src/types/invoice.rs
techraed: done

techraed:
Hey @AnlisUA 
I returned back to the PR. Most of suggestions are implemented. Only the images extensions check is not.

Also I have a question: Shall we validate logo url (from shop meta config)?

**UPD**: images extension check is done. Ready for the review

daemon/src/state.rs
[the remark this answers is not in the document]
techraed: done

PR #257, opened 2026-02-23
feat: add endpoints rotator

### LINKED ISSUE #236: Implement RPC endpoints rotation with circuit breaker
We are already able to put multiple RPC endpoint into config but they are not used now. The idea is to create new structure `RpcEndpointRotator` which will contain all endpoints and their "statuses". RPC endpoint status is another entity which should describe if endpoint is healthy and can be used or it's failed. If it's failed we want to have some timeout before use it again. Each time endpoint fails we want to increase the timeout, if endpoint is already unhealthy, increase timeout for more time (exponential backoff, look for similar implementation for `RetryMeta` placed in `daemon/src/types/common.rs`). If last fail of endpoint happened relatively long time ago (let's say 10 minutes), we can consider that endpoint is heathy.

We can simply wrap this structure into `Arc<RwLock<>` and put into clients. Separate rotators for different chains. When chain client need to be recreated (currently it automatically happen only on transfers subscription failures), we should get new endpoint from rotator and recreate the client with new endpoint. If there are no healthy endpoints at the moment, let's try to return unhealthy, just pick the one which failed the longest time ago.

For periodic endpoint health status check (check if last error happened long enough and we can consider that endpoint is healthy again) we can start another background task which will check it once a minute for all rotators.

die-kreatur:
Something is left to be applied in the future. Now the clients for both Asset Hub and Polygon are created with `from_config` method. Since I added endpoints rotators, config, which also stores endpoints, is not required to be stored in clients structs. For now I couldn't get rid of it because Asset Hub requires `allow_insecure_endpoints` config parameter, while Polygon doesn't, and changing that would affect `BlockChainClient`. I guess it's a job for another task


daemon/src/chain_client/rotator.rs
```diff
@@ -0,0 +1,316 @@
...
+        self.status = RpcEndpointStatus::Healthy;
+        self.attempts = 0;
+        self.last_attempt_at = None;
+        self.next_retry_at = None;
+    }
+}
+
+#[derive(Debug)]
+pub struct RpcEndpointRotator {
+    endpoints: HashMap<String, RpcEndpoint>,
```
AnlisUA: I think it's gonna be better to have `endpoints: Arc<RwLock<...>>` instead of wrapping this structure into `Arc<RwLock<>>`, it will decrease verbosity, we'll get rid of necessity to always make some `rotator.write()` or `rotator.read()` (which potentially can lead to a deadlock), cause we'll call it from within this methods instead and the structure itself can be cloned easily in that case.

Here's another thing about this structure. In some nearest future we're planning to set up our own RPC node in order to save traffic (internal traffic is significantly cheaper then internet one), increase stability (public RPC nodes are not really reliable) and save some response time. With that being said, it might be important to have some priorities for the nodes. Probably the easiest way to achieve that is using `Vec` instead of `HashMap` while it will increase code complexity because of tons of iterators. It's not hard requirement right now but may become so in future. I'll be grateful if you'll try to refactor it to use `Vec` in this PR but if you'll feel that it becomes hard, we can make it in another one.

Also perhaps it's gonna be easier to make a single rotator for all chains. I suppose that under the hood we'll just require all methods to provide us `ChainType` + another cheap `HashMap` call. On the other hand we'll simplify code setup which already looks quite complex in the `main.rs`. Also later we'll definitely want to add some handler which will show endpoints statuses (just read them from rotator), in that case it's also gonna be easier to operate with single rotator vs one-per-chain.

daemon/src/chain_client/asset_hub.rs
```diff
@@ -358,23 +368,28 @@ impl BlockChainClient<AssetHubChainConfig> for AssetHubClient {
...
         config: &crate::configs::ChainConfig,
         asset_info_store: AssetInfoStore<AssetHubChainConfig>,
+        rotator: Arc<RwLock<RpcEndpointRotator>>,
     ) -> Result<Self, ClientError> {
-        AssetHubClient::from_config(config, asset_info_store).await
+        AssetHubClient::from_config(config, asset_info_store, rotator).await
     }
 
     #[instrument(skip(self))]
     async fn recreate(&self) -> Result<Self, ClientError> {
```
AnlisUA: The current approach with marking endpoints unhealthy require us mark it manually every time, which might lead to mistakes, it's quite easy to forget about it. I'd prefer to make it automatically and I suppose that `recreate` methods is a good place for that. The only "bad" thing is that we'll have to store the exact endpoint in every client but on the other hand we'll be able to add this endpoint into our log records and simplify debugging.

And one more thing about it. I suppose it won't be hard to add and use default implementation for this method. Probably we'll have to add separate methods which will return structure fields but they should be dead simple while this method contains can contain some logic (mark endpoint unhealthy at least) and I'd prefer to avoid it's duplication (with ability to make them inconsistent).

AnlisUA (changes requested):

PR #258, opened 2026-02-23
ci: refactor workflows

AnlisUA:


.github/workflows/pr-to-dev.yml
```diff
@@ -0,0 +1,26 @@
+name: PR to dev
+
+on:
+  pull_request:
+    branches-ignore:
+      - main
```
kirushik: This should either be
```suggestion
    branches:
      - develop
```
or (if we want to support nested PRs as a workflow) the workflow (and its file) should be renamed.

.github/workflows/_job-fmt.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Checkout sources
+        uses: actions/checkout@v5
+
+      - name: Install nightly toolchain with rustfmt
+        run: |
+          rustup toolchain install nightly --profile minimal --component rustfmt
+          rustup override set nightly
+
+      - name: Check formatting
+        run: cargo +nightly fmt --all -- --check
```
kirushik: Wouldn't `make cargo-fmt` do essentially the same?
With added bonus of deduplication (so we'd only need to change some additional flags once)


```suggestion
        run: make cargo-fmt
```

.github/workflows/_job-clippy.yml
```diff
@@ -0,0 +1,23 @@
...
+        uses: actions/checkout@v5
+
+      - name: Setup Rust build environment
+        uses: ./.github/actions/setup-rust-build-env
+
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2.7.5
+
+      - name: Run clippy
+        run: cargo clippy --all-targets --all-features
```
kirushik:
```suggestion
        run: make cargo-clippy
```

.github/workflows/_job-release-prepare.yml
```diff
@@ -0,0 +1,64 @@
...
+
+          BODY=$(sed -n "/^## \\[${VERSION}\\]/,/^## \\[/{/^## \\[${VERSION}\\]/d;/^## \\[/d;p;}" CHANGELOG.md)
+
+          {
+            echo "body<<CHANGELOG_EOF"
+            echo "$BODY"
+            echo "CHANGELOG_EOF"
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Create and push tag
```
kirushik: I would really prefer manually created, GPG-signed tags for our releases.
Also: this step is not idempotent as it is now. If our build fails at later step (eg in docker build), it would take a manual tag removal to allow us a second attempt IIUC.

.github/workflows/merge-to-main.yml
```diff
@@ -0,0 +1,46 @@
+name: Merge to main
+
+on:
+  push:
```
kirushik, line 4: Not a fan of "every push to `main` is a full release" workflow. Maybe let's restrict that to the tagged commits on main? Otherwise, say, iterative tinkering with CI, or fixing docs/Readmes wouldn't be possible without a full release pipeline.

.github/actions/setup-sqlite/action.yml
```diff
@@ -0,0 +1,52 @@
+name: Setup SQLite
```
kirushik, line 1: It would be wise to add caching this step, IMO: sqlite is relatively slow-moving target, and this step runs on every test, downloading and compiling it fully from scratch. (Also: potential supply chain attack vector)

.github/workflows/merge-to-main.yml
```diff
@@ -0,0 +1,46 @@
...
+    permissions:
+      contents: write
+
+  docker-build:
+    needs: release-prepare
+    uses: ./.github/workflows/_job-docker-build.yml
+    with:
+      image_tags: >-
+        ghcr.io/${{ github.repository_owner }}/kalatori:${{ needs.release-prepare.outputs.version }},
+        ghcr.io/${{ github.repository_owner }}/kalatori:stable
```
kirushik: Shouldn't we also add `latest` here? We've been publishing `latest` so far, just stopping might break things downstream.
If a more elaborate workflow for tags is implied, it's non-obvious, so please document it.

Dockerfile
```diff
@@ -56,25 +56,34 @@ ENV SQLITE3_INCLUDE_DIR=/usr/local/include
...
+COPY Makefile Makefile
+RUN make install-subxt-cli
+
+# Create source and examples directories
+RUN mkdir -p daemon/src
+RUN mkdir -p client/src
+RUN mkdir -p client/examples
+
+# Copy required Cargo.toml files
+COPY Cargo.toml Cargo.toml
```
kirushik, line 69: You definitely want to copy `Cargo.lock` file inside as well — otherwise your dependency caching is just too shaky to be useful. (And we're committing `Cargo.lock` anyway — which is good and reproduceable; let's make sure our Docker builds are relying on it as well)

kirushik (changes requested):

.github/workflows/_job-release-prepare.yml
AnlisUA: #### About GPGP signing
TBH I don't really like an idea of splitting workflows to `merge-to-main` and `release`. It increases complexity and leaves some space for "forgetting" release or something similar while we intent to automate as many steps as possible. Similar thing with making wrong tag by mistake.

As I understand your idea is to protect the project from "fraud" or "mistake" releases, which was approved in web (without commit signing). I've made a research and seems like we have three (actually two, I don't really like third one) real options:
1. Split workflows as you said, run tests on merge but not release, manually create signed tag and run another flow for release
2. Merge locally and sign merge commit with GPG-key, after push verify that merge commit has been signed by specific key
3. Change trigger from `on push to main` to `on created tag in main`. After merge we'll not run any pipelines but will run only on tag creation

My problems with options 1 and 3 is that they might break the rule "Code in main branch = code running in the `latest` tagged container"

#### About docker build failures
It's not final implementation here, I have a plan to refactor it, wanted to make it in further updates but can (partially or fully) implement in this one. How it should work:
1. Alongside with lints/unit tests we'll build docker container
2. Run integration tests with this container instead of just running app on bare metal
3. Add tags to container and make a dry-run push
4. Create git tag/release/push container to real registry

.github/workflows/merge-to-main.yml
AnlisUA: The core idea is to have `main` branch as a `releases` branch to which we'll merge updates from `dev` branch OR `hotfixes` from respective `hotfix-*` branches. As long as we stay with this idea, ideology "every push to `main` is a full release" actually makes sense IMO, but agree, that in some cases we might need to fix some docs/README and don't make a release from that. Although it sounds more like an edge case then a standard scenario.

I thought about adding some restrictions/requirements of files/folders which should be changed in order to make a release (any source code in our case) while the others (docs/readme) will not require running workflow at all. Does it makes sense for you?

.github/workflows/_job-release-prepare.yml
AnlisUA: For history: discussed in zulip, decided to stick with option 1.

.github/workflows/merge-to-main.yml
AnlisUA: Split flows into post-merge checks and release solves the issue

kirushik (approved):

PR #330, opened 2026-07-28
chore(deps): batch green dependency updates on top of current main

foxcool:
Batches the six green Dependabot PRs into one, **rebuilt from scratch on `main` after #345–#348 landed**.

## Why batch

- **The individual PRs have gone stale.** They were opened in May; resolving the same dependencies today yields newer versions again — `rust_decimal` lands on **1.42.1** (PR #312 proposes 1.42.0, which is still wrong) and `alloy` on **2.1.1** (PR #311 proposes 2.0.5).
- **One review pass instead of six**, for changes that are individually a version number.
- **Granularity is preserved in commits, not PRs.** One commit per dependency, so `git revert` and `git bisect` still work per dependency.

## Contents

| Commit | Change | Supersedes |
|---|---|---|
| `1fa9980` | 5 GitHub Actions SHA pins | #317, #318, #319, #320, #321 |
| `68d9c09` | SHA-pin the three `actions/cache@v5` | — |
| `fbfcce5` | Make the cargo-deny advisories leg blocking | — |
| `2d00a99` | Dependabot grouping config | — |
| `a3568a4` | Declare MSRV once in `[workspace.package]` | — |
| `304a70e` | `cargo update` — 220 package versions | #312 |
| `7d67e96` | alloy 1.8 → 2.0 (resolves to 2.1.1) | #311 |
| `4f30198` | Floor `rust_decimal` at 1.42.1 in both manifests | — |
| `0a675a2` | Pin the websocket reconnect budget — **the one source change** | — |

> [!IMPORTANT]
> **This PR is no longer source-free.** `0a675a2` touches `daemon/src/chain_client/polygon.rs`. Deliberate trade: the alloy 2 bump silently stretches websocket recovery from ~27s to ~195s, and pinning the budget belongs in the same PR as the change that alters it. Everything else is config, docs and manifests.

## Rebuilt, not rebased

The previous version of this branch was cut in July and went `CONFLICTING` once #345–#348 landed. Replaying a 258-package lockfile commit onto a lockfile that had moved four PRs' worth would have been a conflict with no meaningful manual resolution, so the branch was rebuilt from the current `main` instead. Two changes fall out of that:

- **The `ruint` commit is gone.** #347 did the same bump on `main` — same version, same six `ark-*` 0.6.0 lockfile entries, same reasoning. RUSTSEC-2026-0220 is not re-litigated here.
- **The Dependabot config is one commit, not two.** The original added a config and then fixed three defects in it during review. On a rebuild there is no reason to land the broken version first; `2d00a99` is the corrected config, with the three traps written up in its message.

Fresh resolution also moved a few versions on from the July run — `rustls` 0.23.43 rather than 0.23.42, and so on.

## Lockfile

Plain `cargo update`, no manifest changes. 220 of 894 package versions move.

| Crate | From | To |
|---|---|---|
| `rust_decimal` | 1.41.0 | 1.42.1 |
| `reqwest` | 0.13.2 | 0.13.4 |
| `rustls` | 0.23.39 | 0.23.43 |
| `aws-lc-rs` | 1.16.3 | 1.17.3 |
| `zeroize` | 1.8.2 | 1.9.0 |
| `uuid` | 1.23.1 | 1.24.0 |
| `chrono` | 0.4.44 | 0.4.45 |
| `serde` | 1.0.228 | 1.0.229 |
| `spin` | 0.9.8 (yanked) | 0.9.9 |

`tokio`, `axum`, `sqlx` and `subxt` are unchanged — pinned by major in the manifests. **subxt not moving is worth stating**: `AGENTS.md` flags subxt/subxt-cli drift as a project-critical pitfall needing metadata regeneration, and none is triggered. **`hmac`, `hkdf` and `sha2` are unchanged too**, so webhook signatures are byte-identical by construction.

## alloy 1.8 → 2.0: four behavioural changes under one line

No source changes were needed to compile — `polygon.rs`, the ERC-4337/Pimlico path and the ERC-20 helpers all build against alloy 2.x unmodified. **That is exactly what makes this bump easy to under-review.** None of the following touches the API surface:

| # | Delta | Status |
|---|---|---|
| 1 | WebSocket recovery stretches ~27s → ~195s | **fixed here**, `0a675a2` |
| 2 | `alloy-sol-types` 1.6.1 rejects surplus topics, and the decode arm `break`s | fixed in #334 |
| 3 | `reqwest` 0.13.4 attaches the request URL — with the API key — to body errors | fixed in #334 |
| 4 | `sha3` 0.11 / `keccak` 0.2 arrive with nothing pinning expected digests | KATs in #334 |

Three of the four are why **#334 should land before this PR**.

alloy resolves to 2.1.1 rather than the current 2.3.0 because everything from 2.2.0 on requires Rust 1.94.1, above the declared 1.91. That is `resolver = "3"` doing MSRV-aware selection, not a pin — so the matching `ignore` entry in `.github/dependabot.yml` is what keeps Dependabot from proposing it as a "minor" update.

## `0a675a2` — the reconnect budget

Both providers called `WsConnect::new()` and inherited the upstream defaults, 10 retries at 3s. alloy 1.x applies that flat — nine sleeps, **27s**. alloy 2.x keeps both numbers but reads the interval as the base of a capped exponential backoff (cap 30s), so the same configuration becomes 3+6+12+24+30+30+30+30+30 = **195s**. Read from `alloy-pubsub/src/handle.rs` and `service.rs` at both versions.

The length is not the real problem. alloy retries the *same URL*; only `TransfersTracker` can rotate endpoints, by reacting to a dead stream with `recreate()`. So the whole backoff is time that cannot be spent moving off a bad node — the failure in #333. And this module already enforces `WS_MESSAGES_TIMEOUT_DURATION = 10s`, so at the defaults alloy kept retrying for 195s while our own supervision had declared the subscription dead after 10. **The two layers were working against each other**, which was true even before the bump.

Now stated rather than inherited: 3 retries from a 2s base — 6s under 2.x, 4s under 1.x, both inside our 10s.

`ws_reconnect_budget_stays_within_our_own_timeout` pins that relationship, recomputing the worst case from the constants using alloy 2.x's own formula. Duplicating the formula is deliberate — it records the assumption the constants were chosen against, in the one place that fails when upstream changes it again, which upstream has now done once. Mutation-checked: restoring the upstream defaults fails it with `left: 195s, right: 6s`.

## Correction carried over from review

@kirushik flagged that `rust_decimal` 1.42.1 widens `FromStr` to accept scientific notation, so a provider response could newly parse and be silently rounded. **The `FromStr` half is right; the consequence for this repo is not.** Measured against `=1.41.0` and `=1.42.1`, the serde path is unchanged in both — `visit_str` already did `from_str(v).or_else(|_| from_scientific(v))`, byte-identical code. Every `Decimal` here arrives through serde; the only direct `from_str` formats its own `U256`.

The recommendation survives with a worse justification — scientific notation has been accepted at every money boundary all along, `main` included — and is filed as **#344** rather than fixed here.

## Deliberately not included

#310 (subxt-signer 0.50), #313 (hmac 0.13) and #314 (hkdf 0.13). All three are breaking-API migrations, not version bumps, and Dependabot cannot migrate the call sites.

- **#313 and #314 must move together** — `hkdf` depends on `hmac`, so merging either alone puts two `hmac` versions in the tree. `client/examples/generate_hmac_test_vectors.rs` should be used to confirm webhook signatures stay byte-identical.
- **#310** requires a coordinated bump of `subxt`, `subxt-signer` and `subxt-cli` plus metadata regeneration — the pitfall documented in `AGENTS.md`.

Dependabot should close the superseded PRs on its own once this lands.

## Verification

| Check | Result |
|---|---|
| `make cargo-check` (all targets, all features) | pass |
| `make cargo-clippy` (`-D warnings`) | pass |
| `make cargo-deny` | pass — advisories, bans, licenses, sources |
| `make cargo-test` (nextest) | **249/249 passed** |
| Reconnect-budget test, mutation check | fails with `left: 195s` on upstream defaults |

Line references in `4f30198`'s message were re-verified against the post-#348 tree rather than carried over.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PqZWKcMgQuLCeSjbmFELKK


daemon/Cargo.toml
```diff
@@ -67,7 +67,7 @@ subxt = "0.44"
 subxt-signer = "0.44"
 
 # Polygon chain
-alloy = { version = "1.8", features = [
+alloy = { version = "2.0", features = [
```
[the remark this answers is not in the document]
foxcool: Not taking this one, because the mechanism it guards against is already handled — but it points at a real gap next to it.

**Why an upper bound isn't needed here.** The workspace sets `resolver = "3"` (`Cargo.toml:7`), which turns on MSRV-aware dependency resolution. Cargo reads `rust-version` from each candidate version and prefers ones that fit. That is exactly what happened while preparing this PR:

```
Updating alloy v1.8.3 -> v2.1.1 (available: v2.2.0, requires Rust 1.94.1)
```

alloy 2.2.0 was available, Cargo saw it, rejected it against the declared `rust-version = "1.91"`, and said so. A future `cargo update` behaves the same way.

An explicit upper bound would also be inconsistent with the rest of the manifest — `subxt = "0.44"`, `sqlx = "0.8"`, `tokio = "1"`, `serde = "1"` are all open caret requirements. For a binary crate the lockfile is the pin; narrowing the manifest range duplicates the resolver's job and has to be revisited on every legitimate upgrade.

**The real gap.** Nothing verifies that the declared MSRV is true. `rust-toolchain.toml` pins 1.95.0 and every CI job builds with it, so if someone adds code that needs a newer Rust, `rust-version = "1.91"` silently becomes a false claim — Cargo only enforces MSRV on *consumers* (it refuses to build with an older toolchain), never on the code being written. The same blind spot is visible in the docs drift: `AGENTS.md` still says MSRV 1.88 while the manifest says 1.91.

Related: `client/Cargo.toml` declares no `rust-version` at all, even though `kalatori-client` is the crate consumed by other projects, where MSRV is part of the public contract rather than a courtesy.

Filed as #331 — a CI job building with the declared MSRV, plus `rust-version` for the client crate. Out of scope for a dependency batch, so not fixed here.

kirushik (commented):
## Review: dependency batch

Reviewed the full lockfile delta (885 crate-versions, ~190 moved, 30 semver-major) against the changelogs and, where it mattered, against the vendored crate sources rather than the release notes. Since the PR touches no `.rs` files, the entire risk surface is *silent* behavioural change in dependencies — so that is what I went looking for.

**Verdict: sound work, and I'd merge it.** The claims in the description hold up. Nothing here loses funds. But three items deserve action, and one thing in this PR is much better than the description gives it credit for.

---

### Description claims — verified

| Claim | Verdict |
|---|---|
| 5 action SHAs copied verbatim from the Dependabot PRs, not re-resolved from tags | ✅ All 5 resolve to exactly the claimed tag in the claimed repo. `cargo-deny-action` v2.0.19 and `codecov-action` v6.0.1 are *annotated* tags — dereferenced their tag objects, they land on the pinned commits. The other three are lightweight tags pointing straight at them. |
| alloy 2.x needed no source changes | ✅ Semantically true, with two caveats below |
| alloy 2.2.0 requires Rust 1.94.1, so 2.1.1 is the ceiling | ✅ Confirmed against crates.io metadata |
| #310/#313/#314 are breaking migrations, correctly excluded | ✅ And excluding them leaves no inconsistent crypto stack — webhook HMAC still resolves to `hmac 0.12.1` + `sha2 0.10.9`. The new `sha2 0.11.0` enters only via `lzma-rust2`. |
| MSRV drift in docs | ✅ Real. Actual floor of the new lockfile is **1.91**, measured across all 885 crate-versions. The 1.88 in the docs was already wrong before this PR. |

**The fund-critical paths are byte-identical, and I checked that at the source level rather than trusting green CI:**

- `alloy-sol-types/src/eip712.rs` is unchanged between 1.5.7 and 1.6.1 → the EIP-2612 permit digest and the ERC-4337 `PackedUserOperation` hash are identical.
- `alloy-eip7702` never moved off 0.6.3 → `Authorization::signature_hash()` / `into_signed()` unchanged.
- `MnemonicBuilder` still calls the same `mnemonic.derive_key(path, password)` → **derived Polygon deposit addresses are unchanged.** This was the catastrophic case; it's clean.
- `ruint`'s and `alloy-primitives`' serde/Display are untouched → Pimlico JSON-RPC payloads, `Text<Address>` columns and persisted user-op blobs keep their exact format.

---

## Action items

### 1. WS reconnect backoff grew ~7×, and nothing bounds it — `medium-high`

alloy-pubsub 2.x replaced its flat 3s retry with capped exponential backoff (v2.0.2, *"[pubsub] Back off reconnect retries (#3919)"*). With `WsConnect` defaults (`max_retries=10`, `retry_interval=3s`) the 9 sleeps go from `9×3s = 27s` to `3+6+12+24+30+30+30+30+30 = 195s`.

Both connectors are constructed bare, with no override:
- [`polygon.rs:349`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain_client/polygon.rs#L349)
- [`polygon.rs:391`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain_client/polygon.rs#L391)

**Why this matters here specifically:** endpoint failover depends on the provider *failing* so [`transfer_tracker.rs:257`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain/transfer_tracker.rs#L257) falls into `recreate()` and rotates to a different RPC endpoint. A bad endpoint now holds the tracker ~195s instead of ~27s before rotation is even *allowed*. Compounding it, there is **no `tokio::time::timeout` anywhere** in `daemon/src/chain/` or `daemon/src/chain_client/` — only two TODO comments. And [`WS_MESSAGES_TIMEOUT_DURATION`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain_client/polygon.rs#L87) (10s) is now *shorter* than alloy's 30s backoff cap, so the tracker tears down and re-requests a subscription that then sits queued behind the backoff.

**Fix** — keep failover as *your* mechanism, not alloy's:

```rust
let ws_connect = WsConnect::new(&endpoint)
    .with_max_retries(3)
    .with_retry_interval(Duration::from_secs(2));
```

The missing timeouts are pre-existing, but this bump makes them materially worse — worth a follow-up issue.

---

### 2. The Etherscan API key can now reach a panic message, and the daemon aborts — `medium`

reqwest 0.13.4 refactored `bytes()`/`text()`/`json()` through a new `do_bytes()` that attaches the request URL to body/decode errors (`Err(err) => Err(crate::error::decode(err).with_url(*self.url))`), and `reqwest::Error`'s `Debug` renders that URL.

- [`etherscan_client.rs:101-103`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/etherscan_client.rs#L101-L103) — `.text().await.unwrap()`
- [`etherscan_client/types.rs:46-47`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/etherscan_client/types.rs#L46-L47) — the key is a query param, `apikey`

Chain: a mid-body reset (or the 60s request timeout firing during body read) → `unwrap()` formats the error with `{:?}` → the payload contains `...&apikey=<SECRET>` → `PrettyPanic` prints the payload verbatim → [`panic = "abort"`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/Cargo.toml#L145) kills the daemon.

This is a *widening*, not a new class — the existing `tracing::warn!(error.source = ?e, ...)` just above already leaked the same URL for `send()` errors, because reqwest 0.13.2 already did `with_url` on request errors. But `.unwrap()` turns it from a log line into a crash.

**Fix** — the `From<reqwest::Error>` impl at `etherscan_client.rs:44` already discards the value, so this is a one-character change:

```rust
.text().await?   // was: .text().await.unwrap()
```

Longer term, move the key out of the query string into a header, and never format a reqwest error from this client with `{:?}`/`{}` — log `e.is_timeout()` / `e.status()` instead. A test asserting no log record from this client contains the configured key would lock it down.

---

### 3. The hold at alloy 2.1.1 is unenforceable — and the new grouping config defeats it — `medium`

This is the one that will bite silently. Three gaps line up:

1. [`daemon/Cargo.toml:70`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/Cargo.toml#L70) is now `alloy = { version = "2.0", ... }`, i.e. `^2.0`, which permits 2.2.0. No `[patch]`, no `ignore` entry.
2. Dependabot classifies `2.1.1 → 2.2.0` as **minor** (`Cargo::Version.update_type` → `standard_semver_type` → to_minor > from_minor). The new `cargo-minor-patch` group declares `update-types: ["minor", "patch"]`, so it *will* fold the MSRV-breaking bump into next week's batch of trivia.
3. Nothing checks the MSRV. There is no MSRV job in `.github/workflows/` — every job inherits `rust-toolchain.toml`'s `channel = "1.95.0"`. Since 1.94.1 ≤ 1.95.0, it compiles, CI stays green, and [`rust-version = "1.91"`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/Cargo.toml#L8) quietly becomes false.

> On the Copilot thread about an upper bound: your `resolver = "3"` rebuttal is correct as far as it goes — MSRV-aware resolution is real and it *is* what produced `available: v2.2.0, requires Rust 1.94.1`. But it's a *preference*, not a hard constraint, and it doesn't survive Dependabot deciding to bump the manifest floor. That's the gap you yourself pointed at in the same comment.

**Pick one and make it explicit** (a+b is the honest combination):

```yaml
# .github/dependabot.yml, cargo entry
ignore:
  - dependency-name: "alloy"
    versions: [">= 2.2.0"]   # 2.2.0 requires Rust 1.94.1 > our rust-version = "1.91"
```
```bash
# or an MSRV job — install the toolchain explicitly to override rust-toolchain.toml
cargo +1.91 check --workspace --all-features
```

…or decide 1.91 isn't a real constraint, drop it, and let alloy float.

---

## Smaller items

**One bad log kills the whole transfer subscription** — `medium`. alloy-sol-types 1.6 now *rejects extra event topics* (`alloy-rs/core` 1.6.1: *"Reject extra event topics during decoding (#1112)"*). At [`polygon.rs:876-881`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain_client/polygon.rs#L876-L881) a single undecodable log `break`s the `async_stream`, ending tracking for *every* asset, not just the offending one. Same amplification on the `log_to_transfer` arm at [`polygon.rs:872`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain_client/polygon.rs#L872). Canonical USDC always emits 3 topics so the default deployment is fine, but an operator adding a non-standard ERC-20 (ERC-721-ish, ERC-404, re-emitting proxy) to `asset_ids` would see the subscription drop-and-recreate in a loop. **`break` → `continue` is one character and strictly safer**, and it also de-amplifies the pre-existing `abi_decode_data` failure case.

**aws-lc-sys 0.43 auto-detects a system AWS-LC** — `medium`. 0.43.0 adds `builder/system_detect.rs` and a `[target.'cfg(unix)'.build-dependencies] pkg-config`, probing `OPENSSL_DIR` → `OPENSSL_LIB_DIR` → pkg-config `["openssl", "aws-lc", "libcrypto", "libcrypto-awslc"]`. It is **on by default**; only `AWS_LC_SYS_USE_SYSTEM=0` disables it. Since [`main.rs:197`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/main.rs#L197) installs `aws_lc_rs::default_provider()` process-wide, every TLS connection (RPC, Pimlico, Etherscan, webhooks) is terminated by whatever AWS-LC got linked — which may not be the copy pinned by `[package.metadata.aws-lc-sys] commit-hash`. The Dockerfile installs `pkg-config` and `libssl-dev` so the probe *runs*; debian:bookworm ships no aws-lc so it falls back to source today. That's luck, not a guarantee. **Set `AWS_LC_SYS_USE_SYSTEM=0` in the Dockerfile builder stage** — zero behaviour change today, reproducible crypto backend tomorrow.

**Keccak moved onto a rewritten backend** — `low-medium`, and cheap to neutralise. `alloy-primitives` hashes via `sha3::Keccak256` (no `asm-keccak`/`tiny-keccak`/`native-keccak` feature is on — verified with `cargo tree -i`), and this PR moves that onto `sha3 0.11.0` → `keccak 0.2.0`, whose changelog reads *"AArch64 ASM backend is re-implemented using intrinsics"*. Padding and rate constants are provably unchanged (`KECCAK_PAD = 0x01`, `U136`/`U32`), so x86_64 should be bit-identical — this is **not** the Keccak-vs-SHA3 padding footgun. But aarch64 selects new code at runtime via `cpufeatures`, and this path computes permit digests, userOp hashes, and *address derivation* ([`keyring.rs:296`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain_client/keyring.rs#L296)). **A known-answer test costs microseconds and converts this entire class of silent backend swap into a hard failure:**

```rust
assert_eq!(
    keccak256(b""),
    b256!("c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"),
);
// plus one fixed EIP-712 permit digest and one PackedUserOperation hash
// captured before this bump
```

**`actions/cache@v5` is an unpinned mutable tag** — `medium` (CI supply chain). Three occurrences: [`setup-rust-build-env/action.yml:10`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/.github/actions/setup-rust-build-env/action.yml#L10), [`:28`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/.github/actions/setup-rust-build-env/action.yml#L28), [`setup-sqlite/action.yml:9`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/.github/actions/setup-sqlite/action.yml#L9). Every `uses:` under `.github/workflows/` is SHA-pinned, so these are the exception that contradicts the PR's own *"Action pins are SHA-based and reviewed as a set"*. Worse, the new `github-actions` group will **never** surface them: dependabot-core's `file_fetcher.rb` only reads `/action.yml`, `/action.yaml` and `.github/workflows/` when `directory: "/"`. These composite actions run in the job holding `secrets.CODECOV_TOKEN`. Fix: SHA-pin them, and use `directories: ["/", "/.github/actions/*"]` so the group actually covers them.

**cargo-deny's advisories leg can't fail** — `medium`. [`_job-cargo-deny.yml:15`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/.github/workflows/_job-cargo-deny.yml#L15) sets `continue-on-error: ${{ matrix.checks == 'advisories' }}`. So a ~200-crate bump in money-handling software merges with **zero** advisory gating — and `deny.toml`'s `multiple-versions = "warn"` means the other leg won't fail either. Combined with a config explicitly designed to make these PRs land faster, advisory review is now entirely manual on a diff nobody reads line by line. Drop `continue-on-error` and use `[advisories] ignore` for known-accepted findings — that's what the ignore list is for. If flaky advisory-db fetches were the original motivation, note that cargo-deny 0.19.6 (shipped *in this PR's* action bump) specifically hardened advisory-db cloning.

**MSRV docs** — `low`. [`AGENTS.md:3`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/AGENTS.md#L3) and [`docs/conventions.md:5`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/docs/conventions.md#L5) say 1.88; the real floor is 1.91. Also [`client/Cargo.toml`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/client/Cargo.toml) declares no `rust-version` at all — for a crate intended for publication that's a gap in its own right. Consider hoisting `rust-version` to `[workspace.package]` so there's one value.

**`deny.toml` comment is now stale** — `low`. [`deny.toml:7-9`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/deny.toml#L7-L9) justifies the `RUSTSEC-2026-0173` ignore with *"transitive dep via alloy-sol-macro; no safe upgrade available"*. After this PR that's wrong in a good way: alloy-sol-macro 1.6.1 migrated to the maintained `proc-macro-error3` fork (which is why `proc-macro-error3`/`-attr3` show up as new crates). The ignore is still needed — but now for `subxt-macro 0.44.3`, and it can be dropped once subxt migrates. Worth rewording so the next person looks in the right place.

**github-actions group batches majors** — `low`. With no `update-types` key, dependabot-core's `semver_rules_allow_grouping?` returns `true` unconditionally, so the `patterns: ["*"]` group batches everything. That directly contradicts the cargo group's own comment two lines above (*"Major updates stay ungrouped"*), and the docker actions in `_job-docker-build.yml` — the job that publishes the release image with `packages: write` — are exactly the class whose defaults shift across majors. Either mirror the cargo group's `update-types`, or amend the comment to say batching majors is deliberate here.

**Behaviour deltas worth knowing about, no action needed** — `low`. hyper 1.11 now kills keep-alive and strips `Content-Length` on HTTP/1 requests carrying both CL and Transfer-Encoding (RFC 9112 smuggling hardening — strictly desirable; you may just see connection-reuse metrics shift). http 1.4.1/1.4.2 rewrote `PathAndQuery` parsing to reject DEL bytes and non-`/`-leading paths, so a narrow class of malformed request-targets now returns a bare hyper 400 instead of reaching axum's router — relevant only if a test or monitor asserts on the JSON-shaped 404 body. I specifically checked that the stricter parser **cannot** induce the `.parse().unwrap()` in `NormalizePathLayer` to panic (which under `panic = "abort"` would have been serious): it always re-prefixes `/`, and every byte came from an already-accepted path.

---

## The best thing in this PR isn't in the description

**`rust_decimal` 1.42.1 fixes a live money-math bug that is on `main` right now.**

1.41.0 has a broken borrow-propagation stop condition in `unaligned_add`'s subtract path — `if *part > 0 { break; }` where `wrapping_sub(1)` on `0` yields `u32::MAX`, dropping the borrow. Upstream fix: [paupino/rust-decimal#804](https://github.com/paupino/rust-decimal/pull/804), changing it to `if *part != u32::MAX`. I compiled both versions and ran them side by side:

| Expression | 1.41.0 (on `main`) | 1.42.1 (this PR) |
|---|---|---|
| `34028236693 - 7.0000000000000000000000000000` | `68056473378.09…` ❌ | `34028236686` ✅ |
| `34028236701 - 7.0000000000000000000000000000` | `1.906153653662539257` ❌ | `34028236694` ✅ |

It reaches real code: [`invoice.rs:99`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/types/invoice.rs#L99) (unfilled amount), [`transactions_recorder.rs:175`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain/transactions_recorder.rs#L175) (refund size), [`transactions_recorder.rs:228`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/daemon/src/chain/transactions_recorder.rs#L228) (paid/underpaid classification).

Reachability is *extremely* narrow — it needs the rescaled operand to land at ≥ 2¹²⁸ with bits 96..128 exactly 0 or 1 — and 1.8M randomly generated payment-shaped subtractions plus 1.12M values swept across the 2⁹⁴..2¹³¹ band found zero divergences outside the hand-crafted case. So production almost certainly never hit it. But **this is a correctness improvement, not a risk**, and the description frames it as a routine version number. `Display` and `serde` are byte-identical between the two versions (`diff -rq` over both `src/` trees), so no wire-format or `Text<Decimal>` change.

**Two follow-ups from it:**
- 1.41.0 and 1.42.0 were **never yanked**. [`client/Cargo.toml:21`](https://github.com/Kalapaja/Kalatori/blob/ebd26381d976cd224139f88090bc9419d776439c/client/Cargo.toml#L21) is `rust_decimal = "1"` and ships no lockfile to its consumers, so a merchant depending on `kalatori-client` can resolve the broken arithmetic. **Raise the floor to `1.42.1` in both manifests** — the same one-line move this PR already makes for alloy.
- New in 1.42.x: multiplication that underflows to zero can yield a **negative zero** that prints/serialises with a leading minus but compares `Equal` to zero and does not round-trip through `Text<Decimal>`. Not reachable today — your only two production `Decimal` multiplications (`polygon.rs:283`, `:294`) take non-negative operands — but worth a line in `docs/conventions.md` next to the DAO notes, since amount columns are stored as text.

Also positive: `anyhow` 1.0.104 clears **RUSTSEC-2026-0190** and anyhow *is* in the built graph; `codecov-action` v6.0.1 is a template-injection fix (VULN-1652) for the job that handles `CODECOV_TOKEN`; `ed25519-compact` 2.3.1 fixes a genuine zeroize-a-stack-copy bug (dev/test path only for you, but load-bearing if a PASETO signing path is ever added); `zeroize` 1.9.0 replaced `compiler_fence(SeqCst)` with a *stronger* inline-asm barrier, so `Keyring` zeroization is at least as good as before.

---

## Verified clear — so nobody re-litigates these

Each of these looked alarming in the lockfile diff and turned out to be a non-issue on inspection:

- **`secp256k1 0.31.1` and `fixed-cache` appearing under `alloy-primitives`** — lockfile artifacts of weak-feature (`dep?/feat`) edges in 1.6.1's `std` feature. Cargo records the package without activating it. No C libsecp256k1 build, no 16 MiB global keccak cache, no k256→libsecp256k1 recovery-backend swap.
- **`rustls-platform-verifier 0.7.0`** (semver-major) — the major is Android-only. Full source diff of the Linux path (`verification/others.rs`) is three `cfg_attr(docsrs)` attributes and a reworded comment.
- **`jni 0.22.4` / `jni-macros` / `simd_cesu8`** — reached only via `rustls-platform-verifier`'s `cfg(target_os = "android")` and `docsrs` feature. Never compiled for the Linux server build. Same reasoning clears `alloy-signer-aws/gcp/ledger/turnkey`.
- **`alloy-ens`** — `optional = true`, reachable only via the `ens` feature, which you don't enable. `contract` references it weakly as `alloy-ens?/contract`. Never compiled; no ENS resolution path exists.
- **`bitcoin_hashes 0.14.101` / `bitcoin-io 0.1.101`** — legitimate rust-bitcoin `0.14.x` maintenance line (the repo branch resolves; crates.io metadata points at rust-bitcoin), not a typosquat.
- **`jsonrpsee 0.24.11`** — the 0.24.10..0.24.11 range is exactly one functional commit, touching only *server-side* subscription code. Cannot interact with the `TransfersTracker` reconnect fix from #327.
- **`config` 0.15.25 / `yaml-rust2` 0.11 (major)** — the YAML parser is never invoked; every config filename ends in `.json` so `config` infers the JSON format. The `env.rs` change is gated behind `with_convert_case`, which you never call.
- **`serde_json` 1.0.151 + `zmij` 1.0.23** (a big SIMD rewrite) and **`uuid` 1.24.0** — 500k-sample differential tests each, byte-identical output in every case. HMAC-signed webhook bodies and insta snapshots are safe.
- **`chrono` 0.4.45** — the whole delta is 28 lines on the `Local` timezone path. You use `DateTime<Utc>` exclusively. **`time` 0.3.54** isn't on any Kalatori code path at all.
- **`tokio` is unbumped** (1.47.1 in both lockfiles); `quinn` and `io-uring` aren't compiled in; `tokio-util`'s `CancellationToken`/`TaskTracker` changes are purely additive, so shutdown semantics are intact.
- **tower-http 0.6.11** — CORS is enabled but unused; `request_id` diff is test-only; the failure-classification and ServeDir-405 changes need config you don't set.
- **The 17 removed crates** are wasm-bindgen toolchain fallout (`wit-*`, `wasm-encoder`, `prettyplease`). `proptest` is intact at 1.11.0 — it was never yours, it's an optional feature of `alloy-primitives`/`ruint`. Your dev-deps are untouched.
- **The ~14 new duplicate majors** don't trip `bans` — `deny.toml` has `multiple-versions = "warn"`, no deny list, no skip list. Build time and binary size grow; nothing gates.
- **`webpki-roots` 1.0.9** drops `Atos TrustedRoot 2011` and `SecureSign Root CA12`. Affects only the Polygon WS path (reqwest 0.13.4 uses `rustls-platform-verifier` → the OS store, not webpki-roots). No mainstream RPC provider uses either CA, but operators on private/regional endpoints should confirm their WS endpoint still connects.

One pre-existing thing surfaced along the way, not caused by this PR: `tracing-loki`'s default features pull **a second TLS stack** — reqwest 0.12.28 + `native-tls` + OpenSSL — alongside the rustls/aws-lc stack every money path uses. A payment daemon carrying a second C TLS implementation purely for log export is worth a follow-up: `tracing-loki = { version = "0.2", default-features = false, features = ["rustls", ...] }`, then check `cargo tree -i native-tls` comes back empty.

---

One process note, take it or leave it: this PR adds a rule saying majors *"deserve their own review"*, while itself carrying the major bump of the crate that signs Polygon transactions alongside ~200 patch bumps. The grouping config is correct and worth keeping — the point is just that the incentive it creates runs toward rubber-stamping the next batch. Splitting the alloy commit out isn't worth redoing now that this is green and reviewed; I'd rather see the `ignore` entry and the MSRV job land, since those are what actually enforce the property.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code) — lockfile delta analysed against vendored crate sources; differential tests compiled and run for `rust_decimal`, `serde_json`/`zmij` and `uuid`; all 5 action SHAs verified against the GitHub git-refs API.</sub>

daemon/Cargo.toml
```diff
@@ -67,7 +69,7 @@ subxt = "0.44"
...
```
kirushik, line 72: **This one line carries four behavioural changes that don't appear anywhere in the diff.** None of them break the API surface, which is exactly why they're easy to miss — the code compiles unchanged and means something different.

**1. WebSocket recovery stretches from ~27 s to ~195 s.** ⚠️ *highest impact*

Both Polygon providers rely on upstream `WsConnect` defaults, and alloy 2 changed the reconnect algorithm from fixed delays to **capped exponential backoff**. Same code, same call sites, ~7× longer outage.

> A transient Polygon websocket blip now leaves live payment tracking dark for **over three minutes** instead of under thirty seconds. Nothing logs "we changed our recovery budget", because from the daemon's perspective nothing changed.

This compounds with something the grounding pass found separately: alloy transparently reconnects **and re-subscribes**, so the daemon never learns a gap happened — **any `Transfer` emitted while disconnected is silently lost**, and now there's ~7× more window in which to lose one.

*Suggested:* set an explicit, reviewed retry budget on both `WsConnect` values rather than inheriting upstream defaults (they will drift again), and add a test asserting reconnect timing. Longer term, backfill on reconnect — see [`polygon.rs:349`](https://github.com/Kalapaja/Kalatori/blob/b553c3b2d0258bc59bf12cc09e7366b9c9c1132d/daemon/src/chain_client/polygon.rs#L349).

**2. `alloy-sol-types` 1.6.1 rejects surplus topics — and `main` tears down the subscription when decoding fails.**

At [`polygon.rs:881`](https://github.com/Kalapaja/Kalatori/blob/b553c3b2d0258bc59bf12cc09e7366b9c9c1132d/daemon/src/chain_client/polygon.rs#L881) the decode-error arm currently `break`s. So a single malformed or non-standard `Transfer`-like event stops tracking **for every configured Polygon asset**, not just the offending one. This bump makes that path more reachable.

**#334 already fixes this** (`break` → `continue`). Please land #334 first, or merge them together.

**3. `reqwest` 0.13.2 → 0.13.4 turns a latent leak live.** 0.13.4 attaches the request URL to *body* errors; the Etherscan URL carries the API key as a query parameter, and [`etherscan_client.rs:103`](https://github.com/Kalapaja/Kalatori/blob/b553c3b2d0258bc59bf12cc09e7366b9c9c1132d/daemon/src/etherscan_client.rs#L103) `unwrap()`s that result — so a partial response produces a panic whose payload contains the key. **#334 fixes this too.**

**4. New hash backends, no pinned digests.** This brings in `sha3` 0.11 and `keccak` 0.2 alongside the 0.10/0.1 versions. Nothing in the tree pins an expected keccak256, permit or userOp digest, so an architecture-specific hash regression would produce well-shaped but wrong signed digests with every test still green. **#334 adds exactly those known-answer tests** — third reason to sequence it first.

---

None of this argues against the bump: alloy 2.1.1 is also what clears [RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220) via `ruint` 1.20.0. It argues for **saying these four things out loud in the commit message**, pinning the reconnect budget explicitly, and landing #334 first.

The `resolver = "3"` / MSRV note in the root `Cargo.toml` explaining why alloy stays on 2.1.x is genuinely good practice, by the way — that reasoning is usually lost to folklore.

daemon/Cargo.toml
```diff
@@ -18,7 +18,9 @@ kalatori-client = { path = "../client", features = ["axum-middleware", "sqlx-typ
 
 # Types and data serialization utils
 uuid = { version = "1", features = ["v4", "serde"] }
-rust_decimal = { version = "1", features = ["serde", "macros"] }
+# 1.42.1 is a floor, not a preference: 1.41.0 and 1.42.0 lose the borrow when
+# subtracting rescaled operands (paupino/rust-decimal#804) and are not yanked.
+rust_decimal = { version = "1.42.1", features = ["serde", "macros"] }
```
kirushik, line 23: **The reasoning in this comment is correct and I reproduced it. There's a second behaviour change in 1.42.1 that the comment doesn't mention, and it's the one that touches parsing of untrusted input.**

First, confirming the stated fix — built against `=1.41.0`, `=1.42.0` and `=1.42.1`:

```
34028236693 - 7.0000…  →  68056473378.09…   on 1.41.0  ❌
34028236693 - 7.0000…  →  68056473378.09…   on 1.42.0  ❌
34028236693 - 7.0000…  →  34028236686       on 1.42.1  ✅
```

Matches [paupino/rust-decimal#804](https://github.com/paupino/rust-decimal/pull/804). Worth emphasising that **1.42.0 — the version Dependabot proposed in #312 — is still wrong**, so batching to 1.42.1 wasn't just tidiness, it's the difference between fixing this and not. `Display` and serde output for already-constructed values are byte-identical across all three, so no serialization churn.

**The undocumented half:** 1.42.1 also widens `FromStr`. Strings containing `e`/`E` now fall back to [`from_scientific_lossy`](https://docs.rs/rust_decimal/1.42.1/rust_decimal/struct.Decimal.html) instead of being rejected. That affects `DisplayFromStr` and every explicit `Decimal` parse at a money boundary — e.g. [`clients/swaps/bungee/types.rs:64`](https://github.com/Kalapaja/Kalatori/blob/b553c3b2d0258bc59bf12cc09e7366b9c9c1132d/daemon/src/clients/swaps/bungee/types.rs#L64).

> A swap provider returns `"1.0000000000000000005e18"`. Before: parse error, request rejected. After: parses, **silently rounded**, and the rounded figure flows into a money calculation.

That's a loosening of input validation arriving as a patch bump — the sort of thing that's completely invisible in review unless someone diffs the crate source, which is why it's worth a line in the commit message.

*Suggested:* enforce an explicit money grammar and scale at external boundaries (provider responses, API input, config) rather than relying on `Decimal::from_str` to reject malformed values, and add regression vectors for both scientific notation and the subtraction-borrow case above. The subtraction fix is desirable and reachable from the three sites the PR body cites — it's genuinely worth landing.

Cargo.toml
```diff
@@ -7,6 +7,12 @@ members = [
 resolver = "3"
 
 [workspace.package]
+# Actual floor of the current lockfile, measured across all crate versions in
+# the graph. Every member inherits it, so there is one number to keep honest --
+# including the client, which is consumed as a library and for which MSRV is
+# part of the public contract. Also feeds resolver = "3" MSRV-aware version
+# selection, which is what keeps alloy at 2.1.x (see .github/dependabot.yml).
+rust-version = "1.91"
```
kirushik, line 15: **This is the right fix, and it quietly resolves a three-way contradiction — worth coordinating with #337 so you don't fix it twice.**

Before this PR the repo claimed three different Rust versions:

| Source | Claimed |
|---|---|
| `AGENTS.md`, `CONTRIBUTING.md`, `docs/conventions.md` | MSRV **1.88** |
| `daemon/Cargo.toml` | `rust-version = "1.91"` |
| `rust-toolchain.toml` | pins builds to **1.95.0** |

This PR moves the declaration to `[workspace.package]` and updates all three docs to 1.91, which is **verified correct**: scanning `rust-version` across all 611 packages in the lockfile gives a true floor of exactly 1.91, set by alloy 2.1.1. Inheriting it workspace-wide is the right call, and the comment's point about the client being a published library where MSRV is part of the public contract is exactly right.

Two notes:

1. **#337 (open) also touches this contradiction** and currently leaves it unresolved in the same files. Whichever lands second will conflict. Since this PR actually fixes it, #337's job reduces to "don't reintroduce 1.88" — worth a cross-comment so the two don't fight.
2. **MSRV 1.91 is declared but never tested.** `rust-toolchain.toml` pins 1.95.0, so CI only ever proves the code builds on 1.95 — nothing establishes that 1.91 still works, and the guarantee will rot silently. Also note `client/Cargo.toml` and `tools/cargo-bin/Cargo.toml` declare no `rust-version` at all; if the intent is one honest number for the workspace, they should inherit it too. A cheap `cargo +1.91 check` job would turn the claim into something enforced rather than asserted.

The `resolver = "3"` explanation is a nice touch — that's the kind of reasoning that's normally lost the moment the author moves on.

kirushik (commented):
## Review summary

Reviewed with two independent passes that were deliberately not shown each other's output:

- **Adversarial code pass** — GPT-5.6 (`sol`) via codex at high reasoning effort, working from the source tree and the on-disk crate sources for both old and new versions. Asked one question: *what changed underneath code that still compiles?*
- **Upstream-grounding pass** — Claude Opus with web access, checking every version bump against upstream release notes, source diffs and crates.io.

**They disagreed, and that disagreement is the useful part of this review.** The grounding pass verified the *API surface* exhaustively and correctly concluded that all eight documented alloy 2.0 breaking changes miss this repo — it landed on "merge". The adversarial pass found four **behavioural** deltas that don't touch the API surface at all, and landed on "needs work". I checked the load-bearing claims myself and the second pass is right.

**Verdict: this batch is not behaviour-neutral.** It's still the right batch to land — it is currently the only thing fixing a live advisory — but three of the deltas below deserve a decision before it goes in, and the merge order matters.

<details>
<summary><b>Actual before → after, read from both lockfiles</b> (click to expand — several crates carry two majors, so a naive grep misleads)</summary>

| Crate | `main` | this PR |
|---|---|---|
| `alloy` | 1.8.3 | **2.1.1** |
| `alloy-sol-types` | 1.5.7 | **1.6.1** |
| `reqwest` | 0.12.28, **0.13.2** | 0.12.28, **0.13.4** |
| `sha3` | 0.10.9 | 0.10.9, **0.11.0** |
| `keccak` | 0.1.6 | 0.1.6, **0.2.0** |
| `ruint` | 1.18.0 | **1.20.0** |
| `spin` | 0.9.8 (yanked) | **0.9.9** |
| `rust_decimal` | 1.41.0 | **1.42.1** |
| `subxt` / `subxt-signer` | 0.44.3 | **unchanged** ✅ |
| `hmac` / `hkdf` / `sha2` | 0.12.1 / 0.12.4 / 0.10.9 | **unchanged** ✅ |

</details>

### ✅ Verified good, and worth stating explicitly

- **This is the security fix.** `ruint 1.18.0` is subject to **[RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220)** — *"Uint shift operations: incorrect overflow flags and truncated shift amounts"* — and `ruint` is the big-integer crate behind alloy's `U256`, i.e. squarely on the Polygon money path. Every other open branch still carries 1.18.0 and now fails `cargo-deny (advisories)`; this one passes. It also clears the yanked `spin 0.9.8`. Checked separately: no `U256` caller in this repo uses the shift or checked/overflowing-shift APIs whose behaviour changed, so the exposure was latent rather than active — but it should still land.
- **Webhook signatures cannot have moved.** `hmac`, `hkdf` and `sha2` are all untouched, so the bytes are identical by construction. (Deferring #313/#314 was the right call — see the note on the Dependabot config.)
- **subxt coherence holds.** `subxt` and `subxt-signer` both stay at exactly 0.44.3, `subxt-cli` stays on the compatible 0.44 line, so no metadata regeneration was triggered. Worth confirming out loud because `AGENTS.md` flags this as a project-critical pitfall.
- **The alloy money path is byte-stable.** `coins-bip39`, `coins-bip32`, `k256` and `ecdsa` are *identical* on both sides of the lock, so derived Polygon HD addresses are unchanged; `Signature::as_bytes()` byte 64 is still `27 + y_parity` (it just moved into a `v_byte()` helper between 1.5.7 and 1.6.1). EIP-1559 estimator constants are identical at both tags.
- **`secp256k1 0.31.1` is lockfile-only**, not built — `cargo tree -i secp256k1@0.31.1 --workspace --target all` returns nothing, so no new C `libsecp256k1` enters the build.
- **MSRV claim is honest**: all 611 lockfile packages scanned, the true floor is exactly **1.91**, set by alloy 2.1.1.
- `cargo deny` bans / licenses / sources pass; `rustls` and `zeroize` each remain on a single version.

### 🔗 Merge order: please land #334 first

Three of the deltas this PR introduces are **already fixed in #334**, which is currently framed as independent:

| introduced here | fixed in #334 |
|---|---|
| `reqwest` → 0.13.4 attaches the request URL (containing the Etherscan API key) to body errors, and the current code `unwrap()`s that result | removes the `unwrap`, sanitises the error |
| `alloy-sol-types` → 1.6.1 rejects surplus topics, and `main` **`break`s** the whole subscription on any decode error | `break` → `continue` |
| `sha3` 0.11 / `keccak` 0.2 arrive as new hash backends with no pinned digests | adds keccak / permit / userOp known-answer tests |

**Landing this PR first lights up all three.** Recommended: #334, then this — or both together. (This also explains why `ba658af`'s commit message in #334 describes a leak that doesn't match `main`: it was written against the batched tree.)

### 🧾 Pre-existing, surfaced while reviewing this

- **[`daemon/Cargo.toml:135`](https://github.com/Kalapaja/Kalatori/blob/b553c3b2d0258bc59bf12cc09e7366b9c9c1132d/daemon/Cargo.toml#L135) — the `[profile.*]` tables are in a non-root workspace member, so Cargo silently ignores them and `overflow-checks = true` is not applied.** Confirmed from the real `rustc` invocation, not the manifest. Combined with `clippy::arithmetic_side_effects` having been removed in `ed941bc`, this money daemon currently has neither the lint nor the runtime check it believes it has. Being fixed separately; flagging here because this PR is where MSRV/profile hygiene is already being touched.
- **`tools/webhook-simulator` self-test fails on `main` today** — `node tools/webhook-simulator/bin.js --self-test` → *"#8 FAIL — expected `e8599841e652b41b…`, got `d49ecc13344f9588…`", 1 of 10 tests failed*. Not caused by this PR (reproduced on `main`), but nothing in CI runs it, so it has been failing unobserved.
- **The HMAC "test vectors" file is a generator with no committed fixture** (`client/examples/generate_hmac_test_vectors.rs`), so nothing would have caught a signature change had `hmac`/`hkdf` actually moved. That's the safety net this batch happened not to need — worth building before the next one.
- Two TLS stacks still coexist (OpenSSL via `tracing-loki` alongside rustls) — being addressed in #335.

---

*Review method: two blind passes, worktree-isolated, no tracked file modified. Version claims are read from both lockfiles directly; `rust_decimal` behaviour was reproduced in scratch crates rather than inferred.*

daemon/Cargo.toml
foxcool: All four are now stated out loud, and (1) is fixed here rather than deferred — `42fa1e1`.

**Your numbers reproduce exactly.** Read from `alloy-pubsub/src/handle.rs` and `service.rs` at both pinned versions: defaults are 10 retries at 3s in *both*, but 1.x applies them flat (nine sleeps, 27s) and 2.x reads the interval as the base of a capped exponential backoff, `MAX_RECONNECT_RETRY_INTERVAL = 30s`. So 3+6+12+24+30+30+30+30+30 = **195s**.

**What changed my mind about deferring it.** Chasing the mechanism turned up something neither of us said: this module already enforces `WS_MESSAGES_TIMEOUT_DURATION = 10s`. So at the defaults alloy kept retrying for 195s while our own supervision had declared the subscription dead after 10 — the two layers were working against each other, and that was already true at 27s. It is not only "recovery got 7x slower", it is "the inner budget was never coherent with the outer one". That reads like something to fix where the meaning changes, not after.

Budget is now stated rather than inherited: 3 retries from a 2s base — 6s under 2.x, 4s under 1.x, both inside our 10s.

**The test.** `ws_reconnect_budget_stays_within_our_own_timeout` recomputes the worst case from the constants using alloy 2.x's own formula and asserts it stays under our timeout. Duplicating the formula is on purpose — your "they will drift again" is the whole point, and this puts the assumption in the one place that fails when it drifts. Mutation-checked by restoring the upstream defaults, where it fails with `left: 195s, right: 6s`, i.e. it reproduces your figure from our own code rather than asserting a number of its own.

**One correction I owe you, from the other direction.** #334 said this item "only compiles against alloy 2.x, which arrives with #330". That was wrong — `with_max_retries` / `with_retry_interval` exist on `WsConnect` in alloy-transport-ws **1.8.3** with identical signatures (`src/native.rs:93,100`). It belongs here because #330 changes their *meaning*, not their availability. #334's body is corrected.

**Cost, stated plainly:** this PR is no longer source-free. Flagged at the top of the description. Backfill on reconnect stays with #333 — that is genuinely a different mechanism, not a timing constant.

**On putting the four deltas in `3749580` itself:** they are in `42fa1e1`'s message instead. Rewriting the middle of a ten-commit branch drags seven commits through `Cargo.lock` conflicts on a branch that is currently green, which seemed a poor trade for commit-message placement. Say the word if you would rather have it in the bump commit and I will do the replay.

daemon/Cargo.toml
foxcool: **The `FromStr` half reproduces; the consequence for this repo does not.** Built against `=1.41.0` and `=1.42.1` rather than reasoning about it:

| input | 1.41.0 `from_str` | 1.42.1 `from_str` | serde, **both** |
|---|---|---|---|
| `1e18` | `ERR(unknown character)` | `1000000000000000000` | `1000000000000000000` |
| `1.0000000000000000005e18` | `ERR(unknown character)` | `1000000000000000000.5` | `1000000000000000000.5` |
| `1.5E3` | `ERR(unknown character)` | `1500` | `1500` |

`FromStr` did widen, exactly as you said. But **the serde path is unchanged**: `visit_str` already did `Decimal::from_str(v).or_else(|_| Decimal::from_scientific(v))`, byte-identical in both versions (`src/serde.rs:362-368`).

Every `Decimal` that reaches this daemon from outside arrives through serde — API bodies, provider responses, config. There is no `DisplayFromStr` on any `Decimal` field anywhere, and the only direct `Decimal::from_str` in the tree is `polygon.rs:279` inside `u256_to_decimal`, which formats its own input from a `U256` and never sees external text.

So the scenario as written — *"before: parse error, request rejected. After: parses, silently rounded"* — would have parsed and rounded before too, to the same value. (Also, `clients/swaps/bungee/types.rs:64` is `deadline: i64`; the money fields there are `u128`. Nothing on that path goes through `Decimal`.)

**Your recommendation stands, and the real finding is worse than the one you described.** Scientific notation is accepted at every money boundary and always has been, `main` included — a merchant can post `"amount": "1e18"` today and nothing rejects it. There is no grammar or scale check; whatever `Decimal`'s deserializer tolerates is the contract. Filed as #344 with the measurements, including the point that a grammar check without a scale check leaves half the gap open — `1e30` parses fine and exceeds anything our assets can represent, and `Decimal`'s 96-bit mantissa is a second separate limit.

Not fixed here, because it is not a property of this bump — it is equally true before and after.

The subtraction-borrow fix, which is the actual reason the floor is 1.42.1, is untouched by any of this, and your point that 1.42.0 is still wrong is what makes the floor worth stating.

Cargo.toml
foxcool: Thanks — and one of the two notes was already handled here, which I should point at rather than leave you to re-check.

**Inheritance is done.** `client/Cargo.toml:6` and `tools/cargo-bin/Cargo.toml:6` both carry `rust-version.workspace = true` as of `c2636b6`, alongside `daemon/Cargo.toml:8`. Your reading was accurate for `main`; it is the state this PR leaves behind that has all three inheriting. So there is one number, and it is the measured one.

**MSRV declared but never tested** — agreed, and already tracked as #331, so not adding a `cargo +1.91 check` job here. It is the right fix; it just is not a dependency change, and this branch is the wrong place for a new CI job.

**On #337:** the collision is narrower than it looks. #337 does not touch the MSRV lines at all — it only rewrites the clippy-lints section, so nothing there reintroduces 1.88. Both PRs do edit `AGENTS.md` and `docs/conventions.md`, and the hunks are close enough that whichever lands second may want a trivial rebase, but there is no contradiction to resolve twice. I will cross-link so neither of us re-fixes the other's line.

foxcool:
Review pass done, @kirushik — all four threads answered, one pushed as `42fa1e1`.

## The one thing that changed shape

**The WsConnect budget is fixed here, not deferred.** Your 27s → 195s reproduces exactly, read from `alloy-pubsub` at both pinned versions. But chasing the mechanism turned up something neither of us had said: this module already enforces `WS_MESSAGES_TIMEOUT_DURATION = 10s`, so at the upstream defaults alloy kept retrying a dead URL for 195s while our own supervision had declared the subscription dead after 10 — and alloy retries the *same* URL, so that entire window is time `TransfersTracker::recreate()` cannot use to rotate endpoints.

That reframes it. Not "recovery got 7× slower and should be tuned later" but "the inner budget was never coherent with the outer one, and the bump made the incoherence 7× more expensive". Fixing that in the PR that changes the meaning seemed right; deferring it to #333 did not.

Cost stated plainly: **this PR is no longer source-free.** It is flagged at the top of the description. Everything else is still config, docs and manifests.

## Two corrections, both against things I wrote

**#334 was wrong** to say this fix "only compiles against alloy 2.x". `with_max_retries` / `with_retry_interval` exist on `WsConnect` in alloy-transport-ws 1.8.3 with identical signatures. It belongs here because #330 changes their *meaning*. #334's body is corrected.

**The `rust_decimal` concern does not reach this codebase.** `FromStr` widened exactly as you said — measured, `1e18` errors on 1.41.0 and parses on 1.42.1. But the serde path is byte-identical in both versions: `visit_str` already fell back to `from_scientific`. Every `Decimal` here arrives through serde, and the one direct `from_str` formats its own `U256`. So it parsed and rounded before too.

The recommendation survives with a worse justification: scientific notation is accepted at every money boundary and always has been, `main` included. Filed as **#344**.

## Deferred with a pointer, not silently

- **#331** — MSRV declared but never verified. Your `cargo +1.91 check` job belongs there, not in a dependency branch.
- **#333** — backfill on reconnect. Genuinely a different mechanism from a timing constant.
- **#344** — money grammar and scale at external boundaries.
- **#337** — cross-linked. Narrower collision than it looked: it does not touch the MSRV lines at all, only the clippy-lints section, so nothing reintroduces 1.88.

Already handled and worth pointing at rather than re-checking: `client/Cargo.toml:6` and `tools/cargo-bin/Cargo.toml:6` both inherit `rust-version.workspace = true` as of `c2636b6`.

## Verification

`cargo-check`, `cargo-clippy` (`-D warnings`), `fmt --check`, `cargo-deny` all pass. Tests **208/208** — 207 before, one new.

The new test was mutation-checked rather than assumed: restoring the upstream defaults fails it with `left: 195s, right: 6s`, reproducing your figure from our own code rather than asserting a number of its own.

Version claims read from vendored crate sources at the pinned versions — `alloy-pubsub` 1.8.3 vs 2.1.1, `alloy-transport-ws` 1.8.3 vs 2.1.1, `rust_decimal` 1.41.0 vs 1.42.1, the last also built in scratch crates.

**Merge order unchanged: #334 first, then this.** #334 is green and out of draft.

One open question in the alloy thread: the four deltas are spelled out in `42fa1e1`'s message rather than in `3749580` itself, because rewriting the middle of a ten-commit branch drags seven commits through `Cargo.lock` conflicts. Happy to do the replay if you would rather have them in the bump commit.

foxcool:
Rebuilt from scratch on `main` after #345–#348 landed. Force-pushed; nine commits, `MERGEABLE` again.

**Why rebuilt rather than rebased.** The branch went `CONFLICTING` on the lockfile. Replaying a 258-package `cargo update` onto a lockfile that had moved four PRs' worth is a conflict with no meaningful manual resolution — the only honest answer is to resolve the dependencies again, so that is what this does. Fresh resolution lands slightly newer versions than the July run (`rustls` 0.23.43 rather than 0.23.42, and so on).

Two things improved as a side effect:

- **The `ruint` commit is gone.** #347 did the same bump on `main` — same version, same six `ark-*` 0.6.0 entries, same reasoning. Nothing left to re-litigate here.
- **The Dependabot config is one commit instead of two.** The original landed a config with three defects and then fixed them under review. Rebuilding removed the reason to keep that sequence; `2d00a99` is the corrected config, with all three traps written up in its message.

Everything you flagged is still addressed, and the answers are in the threads above — the rebuild only moved their anchors.

## The four alloy deltas are now stated in commit messages

You asked for them to be said out loud rather than left to the diff. They are in `7d67e96`'s message, in the bump commit itself this time — the earlier objection to putting them there was that rewriting the middle of an eleven-commit branch meant dragging seven commits through lockfile conflicts. Rebuilding made that free, so the question I left open in the thread above is resolved in the way you asked for.

## `rust_decimal` line references re-verified

`4f30198`'s message cites `types/invoice.rs:99` and `chain/transactions_recorder.rs:175,228`. Since #348 moved code around, I re-checked all three against the current tree rather than carrying them over — all still point at the subtractions they claim. Given that stale `file:line` was the one weak spot in the review passes, it seemed worth not reproducing it here.

## Verification

`cargo-check`, `cargo-clippy` (`-D warnings`), `fmt --check`, `cargo-deny` all pass. Tests **249/249**.

The reconnect-budget test is mutation-checked: restoring the upstream defaults fails it with `left: 195s, right: 6s`, reproducing your figure from our own constants.

Merge order unchanged: **#334 first, then this.** #334 is rebased, green and mergeable.

kirushik (approved):
🚢

Issue #333, opened 2026-07-28
WS reconnect silently loses subscription events; missed payments reconciled only at invoice expiry


kirushik:
### Summary

When the WS provider sends a frame `alloy-json-rpc` cannot deserialize (observed in production: a JSON-RPC error object with no `id`, emitted by a free public endpoint ~60s into every connection), `alloy-transport-ws` 1.8.3 `handle_text` returns `Err(())` and tears down the **entire pubsub backend**. `alloy-pubsub` reconnects and re-subscribes automatically, but:

1. **The reconnect is invisible**: every step of `reconnect()` logs at `debug!`, and the `error!("Pubsub service backend error.")` arm is unreachable behind a `biased` `select!`. At INFO the only trace is the deserialize ERROR line itself. Kalatori's own `Subscription` watchdog (10s) never fires because the gap is shorter.
2. **Events in the gap are lost forever on the live path**: `eth_subscribe("logs")` has no backfill. Measured on Polygon (~2s blocks) with a ~0.28s reconnect blind window: roughly 1 in 7 reconnects drops a block's log burst. At 60 reconnects/hour that was ~0.5% of blocks invisible to the affected instance.
3. **The safety net is a day late**: a payment landing in a blind window is only caught by the `BalanceChecker`→Etherscan reconciliation at **invoice expiry** (default 24h, `expiration_detector.rs:170`) — the invoice sits "waiting" and the shop webhook stays unfired until then.

Funds are not at risk; payment confirmation latency is.

### Suggested fixes (independently useful)

- **Block-gap backfill after reconnect**: track the last block seen per subscription; on resubscribe, fetch logs for the gap via `eth_getLogs` before resuming live processing. This closes the loss window for *any* reconnect cause (provider hiccups, network blips), not just this one.
- **Run the reconciliation sweep periodically**, not only at invoice expiry, so a missed event is caught in minutes rather than hours.
- **Fallback visibility**: log a startup WARN when a daemon falls back to the built-in `DEFAULT_POLYGON_ENDPOINTS` (`daemon/src/configs/consts.rs`) instead of configured endpoints — the affected instance ran on a free public provider for three months without a single log line saying so. Consider whether free public providers belong in the compiled-in defaults at all.
[...]

Issue #331, opened 2026-07-28
MSRV is declared but never verified; kalatori-client declares none


foxcool:
## Problem

`daemon/Cargo.toml` declares `rust-version = "1.91"`, but nothing verifies that the declaration is true.

`rust-toolchain.toml` pins the toolchain to 1.95.0, and every CI job builds with it. Cargo enforces MSRV only on the *consuming* side — it refuses to build a crate with a toolchain older than its `rust-version` — but it never checks that the code being written actually compiles on the declared minimum. So the moment someone uses a feature stabilised after 1.91, CI stays green and `rust-version = "1.91"` quietly becomes a false claim. It would surface only when an external consumer tries to build on 1.91 and hits a wall of syntax errors.

The docs drift is a symptom of the same blind spot: `AGENTS.md` still states MSRV 1.88 while the manifest says 1.91. Nobody reconciles the number because nothing validates it.

This is not hypothetical bookkeeping — MSRV actively drives dependency resolution. The workspace uses `resolver = "3"`, so Cargo picks dependency versions that fit the declared minimum. While preparing #330:

```
Updating alloy v1.8.3 -> v2.1.1 (available: v2.2.0, requires Rust 1.94.1)
```

alloy stopped at 2.1.1 precisely because of `rust-version = "1.91"`. A stale or wrong MSRV therefore silently changes which dependency versions the project gets.

## `kalatori-client` declares no MSRV at all

`client/Cargo.toml` has no `rust-version` field. That is the inverse of where it matters most: `kalatori-client` is the crate consumed by other projects, so its MSRV is part of the public contract, whereas for the daemon binary it is a courtesy.

## Suggested fix

1. Add a CI job that builds with the declared minimum rather than the pinned toolchain:

   ```yaml
   - run: rustup toolchain install 1.91 --profile minimal
   - run: cargo +1.91 check --workspace --all-features --all-targets
   ```

   Alternatively `cargo-msrv verify`, which reads `rust-version` instead of hardcoding it in the workflow — that keeps the manifest as the single source of truth.

2. Add `rust-version` to `client/Cargo.toml`, matching or below the daemon's.

3. Update the MSRV figure in `AGENTS.md` (currently 1.88) and add it to `docs/doc-update-triggers.md` so it is refreshed whenever the manifest changes.

## Context

[...]

PR #334, opened 2026-07-30
fix(polygon): address source-level findings from the #330 review

foxcool:
Source-level findings from @kirushik's review of #330, split out so that PR could stay
deps-only. Based on `main`, not on #330 — no file overlap with it.

One commit per finding, so `git revert` still works per decision.

| Commit | Finding | Origin |
|---|---|---|
| `a58aa6d` | `break` → `continue` in the transfer subscription | #330 review, medium |
| `4443d56` | Known-answer tests for keccak256, permit digest, userOp hash | #330 review, low-medium |
| `a34d886` | Etherscan API key reaching the logs, and the `unwrap()` crash | #330 review, medium |
| `2d785b7` | `AWS_LC_SYS_USE_SYSTEM=0` on every build surface | #330 review, medium |
| `fe91d91` | The same key still reaching `Debug` by two other paths | review of this PR |
| `5999ae2` | Provenance for the known-answer vectors | review of this PR |
| `c9148c2` | `ruint` 1.20.0 for RUSTSEC-2026-0220 | see "Why the advisory fix is here" |

## `a58aa6d` — one bad log no longer kills the whole subscription

Both error arms in the `async_stream` loop (`polygon.rs:867,881`) broke out of it, so a single
log that failed to decode or record tore down tracking for **every** asset on that
subscription. The stream ended, the tracker recreated the subscription, and the next bad log
did it again.

This gets sharper with alloy-sol-types 1.6, which rejects logs carrying extra event topics.
Canonical USDC always emits exactly three, so a default deployment is unaffected; an operator
who adds a non-standard ERC-20 to `asset_ids` gets a drop-and-recreate cycle.

**The trade-off is deliberate and worth flagging.** `continue` turns a loud failure into a
skipped event, and without backfill (#333) that event is gone for good. So the processing arm
is raised `warn` → `error` and now logs the transaction hash: the log decoded, so it is one of
our `Transfer` events, and failing to record it means a payment nobody sees until the invoice
expires. The decode arm stays at `debug` — a log we cannot decode is shaped for someone else.

Making this *visible per asset* rather than merely survivable is the obvious next step and is
tracked separately — see "Deliberately not here".

## `4443d56` + `5999ae2` — pinning the hashes, and saying where they came from

`alloy-primitives` hashes through `sha3::Keccak256` — neither `asm-keccak`, `tiny-keccak` nor
`native-keccak` is enabled, confirmed with `cargo tree -i`. On aarch64 that backend picks an
assembly path at runtime via `cpufeatures`, and `keccak` 0.2.0 reimplements it with intrinsics.
A silently substituted backend would change permit digests, userOp hashes and **address
derivation** without failing to compile.

Three tests, microseconds each: `keccak256` against canonical vectors, `build_permit_hash`, and
`compute_user_op_hash` over the existing `default_polygon_unsigned_transaction()` fixture.

`keccak256("")` is `c5d24601…`, not `a7ffc6f8…` — the latter is SHA3-256, which is what a
padding regression looks like.

Expected values are hardcoded on purpose: they must fail on any change of result, including one
that looks like an upgrade. Verified by flipping a byte in each and confirming all three fail.

`5999ae2` records **provenance per constant**, because nothing in the file distinguished a
verified vector from a snapshot of a bug — a KAT generated by the code it guards locks in
whatever that code does wrong. @kirushik verified all three against sources outside this
implementation, and each constant now carries which contract, which method, which chain and
which EntryPoint version. The file also states the rule those comments exist to enforce: a
failing value is re-derived the way its comment says, never refreshed from the new output.

`build_permit_hash` and `compute_user_op_hash` lose their `&self`. Neither ever used it — both
read only their arguments and the module constants — but the receiver meant a test had to
construct a `PolygonClient`, which needs a live websocket. Call sites become `Self::`.

## `a34d886` + `fe91d91` — the Etherscan key

Two defects on the same three lines, with different severities:

**The live one is the `warn!`.** reqwest attaches the request URL to `send()` errors, already at
the pinned **0.13.2** — verified in `client.rs`, which calls `.with_url(...)` — and the API key
is a query parameter. So `error.source = ?e` has been writing the key into the logs on every
connection failure and timeout. It now logs `is_timeout` and `is_connect`. `status` is dropped
rather than kept: `send()` failures are `Kind::Request`, and `Error::status()` answers `Some`
only for `Kind::Status`, so the field was always `None`.

**The second is `.text().await.unwrap()`**, which crashed the daemon on any body or decode
error. At 0.13.2 that payload carries no key: `text()`/`bytes()` map their failure through
`error::decode`, which constructs the error with `url: None`. So today it is a crash, not a
leak — a partial body read aborting a payment daemon under `panic = "abort"`. **That changes
with #330**: reqwest 0.13.4 routes body reads through `do_bytes`, which attaches the request
URL. Verified in both crate sources. This is the strongest argument for the merge order below.

`.text().await?` instead, with an `inspect_err` logging the same safe flags, so Etherscan
flakiness is still diagnosable.

`fe91d91` closes the rest of the surface, because partial redaction is worse than none — it
reads as done. Two structs held the key in plaintext behind a derived `Debug`:
`EtherscanClientConfig` and `GetAccountTokenTransactionsParams` (which carried a `TODO` asking
for exactly this). Both now redact, and both are covered by tests that format the value with a
sentinel key and assert its absence. **This is not a new pattern here** — `BungeeApiConfig` and
`ZeroExApiConfig` already declare `api_key: SecretString`; Etherscan was the one key still on a
bare `String`.

Moving the key into a header is the real fix and needs checking against the Etherscan v2 API;
tracked separately.

## `2d785b7` — AWS-LC

**Forward-looking hygiene, inert today.** The earlier version of this commit claimed the probe
runs on every image build. It does not, at the version this repo locks: `aws-lc-sys` **0.40.0**
has no system-probing code at all — its `builder/` contains no `USE_SYSTEM`, `OPENSSL_DIR`,
`OPENSSL_LIB_DIR` or `pkg_config`. The machinery arrives in **0.43.0** as
`builder/system_detect.rs`, and even there it only adopts a candidate carrying the
`OPENSSL_IS_AWSLC` marker, so Debian's `libssl-dev` was never a candidate. #330 is what brings
0.43.0 in.

It is worth setting early because `main.rs:197` installs `aws_lc_rs::default_provider()`
process-wide, so every TLS connection — RPC, Pimlico, Etherscan, merchant webhooks — terminates
on whatever crypto got linked at build time.

Now set in **`.cargo/config.toml` as well as the Dockerfile**. The Dockerfile alone covers only
the release image: PR CI does not build it, and local and CI cargo builds never see its `ENV`.
After the bump that would let the image vendor AWS-LC while a test build links a detected system
one. Not `force` in `[env]`, so a deliberate override still wins.

## Why the advisory fix is here — `c9148c2`

This branch's `cargo-deny (advisories)` was **stale-green**: it last ran before
[RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220) (`ruint` — incorrect
overflow flags and truncated shift amounts) was published, and the branch carried the vulnerable
`ruint` 1.18.0. Any push turns it red, so the check had to move with the branch.

Cherry-picked from #330 (`b553c3b2`). `cargo update -p ruint --precise 1.20.0` also writes six
`ark-*` 0.6.0 entries into the lockfile — ruint 1.20.0 declares them optional. None enters the
build graph (`cargo tree -i ark-ff@0.6.0 --workspace --target all` prints nothing), and #330
resolves to the same set. `cargo deny check` now passes on this branch: advisories, bans,
licenses, sources.

## 🔗 Merge order — please read before landing either PR

Three of the changes here are **fixes for hazards that #330 introduces**, not for hazards that
exist on `main` today:

| #330 introduces | fixed by, in this PR |
|---|---|
| `reqwest` 0.13.2 → **0.13.4**, which attaches the request URL (containing the API key) to body errors | `a34d886` |
| `alloy-sol-types` 1.5.7 → **1.6.1**, which rejects surplus topics — and `main`'s `break` aborts the whole subscription on any decode error | `a58aa6d` |
| `sha3` 0.11 / `keccak` 0.2 arrive as new hash backends with nothing pinning expected digests | `4443d56` |

So this PR is only *nominally* independent of #330. **Landing #330 first lights up all three.**
Recommended order: **this PR first, then #330** — or both together.

## Deliberately not here

~~**`WsConnect::with_max_retries(3)` / `with_retry_interval(2s)`**~~ — **done in #330** (`42fa1e1`),
where the bump that changes the defaults lives.

**Correction to an earlier version of this section**, which said the change "only compiles
against alloy 2.x, which arrives with #330". That is wrong: `with_max_retries` and
`with_retry_interval` exist on `WsConnect` in alloy-transport-ws **1.8.3** with identical
signatures (`src/native.rs:93,100`), so it would have compiled here too. The reason it belongs
in #330 is that #330 is what changes their meaning — 1.x reads the interval as a flat delay,
2.x as the base of a capped exponential backoff — not that the API is unavailable.

**Per-asset health tracking.** `continue` is strictly better than the old `break`, but a
persistently undecodable asset is now skipped forever while the shared subscription looks
healthy. Surfacing that needs per-asset state and a rate-limited `warn!` — a behaviour design
that interacts with the same backfill work, not a line change to fold in here.

**A Docker build in PR CI.** `_job-docker-build.yml` is only called from `merge-to-main.yml` and
`release.yml`, so a Dockerfile change merges on review alone and is first exercised after it
lands. Real gap, but it is a CI change rather than a review finding.

## Verification

| Check | Result |
|---|---|
| `make cargo-check` (all targets, all features) | pass |
| `make cargo-clippy` (`-D warnings`) | pass |
| `make cargo-deny` | pass — advisories, bans, licenses, sources |
| `make cargo-test` | **212/212** (210 before, 2 new) |
| KAT mutation check | all three fail on a one-byte change |
| Redaction mutation check | reverting the hand-written `Debug` to a derive fails the test with the sentinel visible |

Library claims were read from the vendored crate sources at the pinned versions, not from
release notes: reqwest 0.13.2 vs 0.13.4 (`error.rs`, `async_impl/response.rs`), aws-lc-sys
0.40.0 vs 0.43.0 (`builder/`).

**`2d785b7` is still not verified by anything.** It could not be built locally, and the PR
pipeline does not build the image either. So a Dockerfile change merges on review alone.

`a58aa6d` has no unit test either: both arms live inside `async_stream!` around a live
subscription, and reaching them needs a mock websocket that does not exist yet. It was checked
by reading — `continue` binds to the `loop`, not to the `select!`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PqZWKcMgQuLCeSjbmFELKK


daemon/src/chain_client/polygon.rs
```diff
@@ -865,20 +866,26 @@ impl BlockChainClient<PolygonChainConfig> for PolygonClient {
...
                             Err(e) => {
+                                // One log we cannot decode must not take down tracking for
+                                // every other asset on this subscription.
                                 tracing::debug!(
                                     error = ?e,
-                                    "Failed to decode Transfer event from log"
+                                    "Failed to decode Transfer event from log, skipping it"
                                 );
-                                break
+                                continue
```
kirushik, line 1021: **`continue` is the right call — confirmed against alloy 1.8.3. But it converts a loud failure into a silent one, and nothing downstream learns about it.**

Why the change is correct: `SubscriptionStream<Log>::Item` is `Log`, not `Result`. Lag and deserialisation failures are handled *inside* alloy; only channel closure yields `None`. Both `Err` arms here are per-log conditions, so `break` was tearing down a subscription that was working fine. (Terminal closure still reaches `TransfersTracker`, which backs off, rebuilds the client and resubscribes — verified, 11 lifecycle tests pass.)

The gap: this arm now skips the log at `debug!` level and **no per-asset degradation reaches `TransfersTracker`**. Concrete failure:

> An operator configures a non-standard ERC-20 whose `Transfer` carries an extra indexed topic. `alloy-sol-types` refuses to decode it. Every payment for that asset is skipped, forever — while traffic on the *other* assets keeps the shared subscription looking perfectly healthy. No alert fires, because from the tracker's point of view nothing is wrong.

That's strictly better than the old behaviour (which broke tracking for *all* assets), so please don't read this as "revert". It's the next step.

**Suggested:**
- Track health **per asset**, not per subscription, so a persistently-undecodable asset can be surfaced.
- Rate-limit and raise the level: a *persistent* decode failure is an operator-actionable condition, not `debug!`. One `warn!` per asset per interval, rather than one per log.
- Add a `valid log → undecodable log → valid log` test asserting the stream survives *and* that the skip is observable.
- Per `docs/conventions.md`, these two arms should also carry `error.category = category::CHAIN_CLIENT` and `error.operation` from `daemon/src/utils/logging.rs` and use `error.source` rather than `error` — otherwise the documented production log queries won't match skipped-payment events.

This becomes considerably more load-bearing once #330 lands: `alloy-sol-types` 1.6.1 tightens surplus-topic rejection, so the undecodable-log path gets *more* traffic, not less.

daemon/src/etherscan_client.rs
```diff
@@ -93,14 +93,17 @@ impl EtherscanClient {
             .send()
             .await
             .inspect_err(|e| {
+                // Never format a `reqwest::Error` from this client with `{:?}` or `{}`:
+                // it carries the request URL, and the API key is a query parameter.
                 tracing::warn!(
-                    error.source = ?e,
+                    timeout = e.is_timeout(),
+                    connect = e.is_connect(),
+                    status = ?e.status(),
```
kirushik: **The fix is right, and it's incomplete. Partial redaction is worse than none, because it reads as done.**

What's correct: at the pinned **reqwest 0.13.2**, `send()` errors *do* carry the URL — [`client.rs` attaches it via `.with_url(...)`](https://github.com/seanmonstar/reqwest/blob/v0.13.2/src/client.rs), and `Debug` prints it. The API key is a query parameter, so `error.source = ?e` was a real leak. Removing it is the right call.

Three follow-ups, in severity order:

**1. The key still reaches logs by other paths.** Deriving `Debug` over a struct holding the plaintext key means any `?config`/`?params` anywhere formats it:
- [`configs/types.rs:429`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/configs/types.rs#L429) — `#[derive(Deserialize, Clone, Debug)]` on `EtherscanClientConfig`
- [`etherscan_client/types.rs:37`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/etherscan_client/types.rs#L37) — `#[derive(Debug, Serialize)]` over `api_key: &str`

Suggest `secrecy::SecretString` (or a hand-written `Debug` printing `apikey=***`) plus a test that formats the value with a sentinel key and asserts the sentinel is absent. That test is what makes the guarantee durable.

**2. `status = ?e.status()` is dead here.** `Error::status()` returns `Some` only for `Kind::Status` errors; a `send()` failure is `Kind::Request`, so this field is always `None`. Harmless, but it looks like diagnostics and isn't.

**3. A version-proof alternative exists.** reqwest offers [`Error::without_url()`](https://docs.rs/reqwest/0.13.2/reqwest/struct.Error.html#method.without_url), which strips the URL and lets you keep logging the full error chain. That survives the 0.13.4 change described in the next comment, where *more* error paths gain URLs — a denylist of known-leaky fields will not.

daemon/src/etherscan_client.rs
```diff
@@ -93,14 +93,17 @@ impl EtherscanClient {
...
                     "Etherscan request failed"
                 )
             })?
             .text()
-            .await
-            .unwrap();
+            .await?;
```
kirushik: **Good change — but `ba658af`'s commit message describes a leak that doesn't exist yet on `main`.**

The message says the API key reached "a panic payload". At the pinned **reqwest 0.13.2** that isn't so: `.text()`/`bytes()` map their failure through [`error::decode`](https://github.com/seanmonstar/reqwest/blob/v0.13.2/src/error.rs), which constructs the error with `url: None`. So the old `.unwrap()` could panic, but the payload carried no key.

URL attachment on body errors starts in **0.13.4** (`do_bytes` attaches the request URL). And #330 bumps `reqwest` to exactly 0.13.4.

So the claim isn't wrong so much as **early**: it's true of the tree *after* #330 lands, which is presumably the tree it was written against before being split out. Worth correcting the message so the git archaeology still makes sense in a year — and worth noting that this is the strongest argument for landing this PR **before** #330 rather than after.

Removing the `unwrap()` is independently correct anyway: a partial-body read shouldn't abort a payment daemon.

Dockerfile
```diff
@@ -77,6 +77,15 @@ FROM base AS builder
...
 WORKDIR /usr/src/kalatori
 
+# Build AWS-LC from the vendored sources pinned by aws-lc-sys, never from a
+# system install. Since aws-lc-sys 0.43.0 the build script probes OPENSSL_DIR,
+# OPENSSL_LIB_DIR and pkg-config by default, and this stage installs both
+# pkg-config and libssl-dev, so the probe does run. main.rs installs
+# aws_lc_rs::default_provider() process-wide, which means every TLS connection
+# the daemon makes -- RPC, Pimlico, Etherscan, webhooks -- terminates on
+# whatever got linked here.
+ENV AWS_LC_SYS_USE_SYSTEM=0
```
kirushik, line 93: **This `ENV` is currently inert, and the comment's reasoning is inverted.**

The comment says the probe "does run" — at the version this repo actually pins, it can't. `Cargo.lock` pins **`aws-lc-sys` 0.40.0**, and that version's `builder/` contains **zero** occurrences of `USE_SYSTEM`, `OPENSSL_DIR`, `OPENSSL_LIB_DIR` or `pkg_config`. The system-probing machinery you're defending against [first appears in 0.43.0](https://github.com/aws/aws-lc-rs/blob/main/aws-lc-sys/builder/main.rs). And even there it only adopts a candidate library carrying the `OPENSSL_IS_AWSLC` marker, so Debian's `libssl-dev` could never have been linked regardless.

So today this line changes nothing. That's fine as forward-looking hygiene — but the comment states a false mechanism, and a future reader will act on it.

There's a second-order problem worth fixing at the same time: **the variable is set only in the Dockerfile.** PR CI doesn't build the image, and ordinary local/CI builds never set it. After a future bump to aws-lc-sys ≥ 0.43, Docker would vendor AWS-LC while test and dev builds could link a *detected system* one — divergent crypto backends per build surface, which is exactly the class of thing that produces "works in CI, fails in prod".

**Suggested:**
1. Reword to what's actually true — something like *"aws-lc-sys 0.40.0 has no system-probe; this pins the intent ahead of ≥0.43.0, where the probe is introduced."*
2. Apply the policy to **every** build surface (workspace `.cargo/config.toml` `[env]`, or the CI job env), not just the image.
3. When the dependency does cross 0.43, assert the selected backend in CI rather than trusting the env var.

daemon/src/chain_client/polygon.rs
```diff
@@ -1287,8 +1297,63 @@ impl BlockChainClient<PolygonChainConfig> for PolygonClient {
...
+    // silently substituted backend would change permit digests, user operation
+    // hashes and address derivation without failing to compile. These vectors are
+    // hardcoded on purpose: they must fail on any change of result, including one
+    // that looks like an upgrade.
+
+    /// Canonical Keccak-256 vectors. Note these are Keccak, not SHA3-256 --
+    /// SHA3-256("") is `a7ffc6f8...`, which is what a padding regression looks
+    /// like.
+    #[test]
+    fn keccak256_matches_known_vectors() {
```
kirushik, line 1480: **These vectors are correct — I verified them against the chain, not against the code. Please record that provenance here, because nothing in the file currently distinguishes a verified vector from a snapshot of a bug.**

What was checked, independently of this implementation:

| Constant | Verified against | Result |
|---|---|---|
| userOp hash | live `eth_call` to `getUserOpHash()` on EntryPoint v0.8 [`0x4337084D…F108`](https://polygonscan.com/address/0x4337084d9e255ff0702461cf8895ce9e3b5ff108), Polygon mainnet, with this fixture | `0xadcbc48b…1853` — byte-identical |
| permit domain separator | USDC's on-chain `DOMAIN_SEPARATOR()` | matches `eip712_domain!` exactly |
| permit typehash | USDC's on-chain `PERMIT_TYPEHASH()` | matches the literal exactly |
| keccak256 | from-scratch Keccak-f[1600], cross-checked vs CPython `hashlib.sha3_256` | matches |

Also confirmed: `0x4337084D…F108` is **v0.8**, and the code uses its `("ERC4337", "1")` EIP-712 domain — so this is *not* a [v0.6 hash construction](https://eips.ethereum.org/EIPS/eip-4337) applied to a v0.7+ EntryPoint, which is the classic way this test passes while production userOps get rejected.

Making `build_permit_hash` / `compute_user_op_hash` take no `self` purely so tests can call them is a nice move, and the comments explaining why are exactly right.

**Suggested additions:**
1. **Provenance comment per constant** — one line each: which contract, which method, which chain, which EntryPoint version. Without it, the next person to "refresh" a failing constant from implementation output will silently enshrine a regression, and this file gives them no reason to hesitate.
2. **Boundary cases.** The fixture is non-production-like (empty `callData`, zero-ish gas fields). The packed-field encoding is where EntryPoint versions actually differ — add a case with non-empty `callData`, and max / one-over-max values for the packed gas fields. That last one doubles as a regression test for the `U256::to::<u128>()` panic at [`polygon.rs:582`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/chain_client/polygon.rs#L582).
3. **Document the pattern** in `docs/testing-strategy.md` — `docs/doc-update-triggers.md` requires it for a new test pattern, and "KATs must have independently-sourced expected values" is precisely the convention worth writing down once.

One version note for later: #330 introduces `sha3` 0.11 / `keccak` 0.2 as a *new hash backend*. These tests are what will catch a regression there, which is another reason to land this PR first.

(Minor: the `aarch64` motivation in the test comment doesn't apply here — `keccak 0.1.6`'s ARMv8-SHA3 dispatch is `#[cfg(all(target_arch = "aarch64", feature = "asm"))]`, and `sha3 0.10.9` has `asm` off by default, with CI on x86_64 only. The tests are worth keeping; just the stated reason isn't the real one.)

kirushik (commented):
## Review summary

Reviewed with two independent passes that were deliberately not shown each other's output:

- **Adversarial code pass** — GPT-5.6 (`sol`) via codex at high reasoning effort. Logic, edge cases, security, test integrity. Ran the *real* CI invocation `make cargo-clippy` (`-Dwarnings`, all targets, all features): **passes**. Tests: `4 passed` (polygon), `11 passed` (transfer_tracker).
- **Upstream-grounding pass** — Claude Opus with web access. Every third-party claim checked against the pinned versions, upstream sources, and **live Polygon mainnet calls**. 29 claims checked: 24 confirmed, 3 refuted.

**Verdict: the code is right, three prose claims are wrong, and the API-key redaction is incomplete.** Nothing here is a reject — it's a revision pass.

<details>
<summary><b>Pinned versions this review is grounded in</b> (click to expand)</summary>

| Crate | Version at this PR head |
|---|---|
| `alloy` | 1.8.3 (`alloy-primitives` / `alloy-sol-types` 1.5.7) |
| `reqwest` | 0.12.28 **and 0.13.2** (two majors coexist) |
| `rustls` | 0.23.39 · `aws-lc-rs` 1.16.3 · **`aws-lc-sys` 0.40.0** |
| `sha3` | 0.10.9 · `keccak` 0.1.6 |

Every claim below is about *these* versions, not "latest".
</details>

### ✅ What was verified good

**The known-answer tests are genuinely correct — not tautological.** This was the single most important thing to check, because a KAT generated by the code it guards silently locks in whatever bug that code has. These weren't:

- EntryPoint v0.8's `getUserOpHash()` was called **live on Polygon mainnet** at [`0x4337084D9E255Ff0702461CF8895CE9E3b5Ff108`](https://polygonscan.com/address/0x4337084d9e255ff0702461cf8895ce9e3b5ff108) with this PR's own fixture → `0xadcbc48b…1853`, byte-identical to the committed constant.
- USDC's on-chain `DOMAIN_SEPARATOR()` and `PERMIT_TYPEHASH()` match the `eip712_domain!` block and the literal exactly.
- Keccak was independently recomputed from a from-scratch Keccak-f[1600] implementation (cross-checked against CPython's `hashlib.sha3_256`) in **both** passes.
- `0x4337084D…F108` **is** EntryPoint v0.8, and the code correctly uses its [`("ERC4337", "1")` EIP-712 domain](https://eips.ethereum.org/EIPS/eip-4337) — so there is no v0.6/v0.7 construction mismatch, which was the specific failure mode I went looking for.

**`break` → `continue` is correct.** In alloy 1.8.3, `SubscriptionStream<Log>::Item` is `Log`, not `Result`: lag and deserialisation failures are already swallowed inside alloy, and only channel closure yields `None`. Both `Err` arms here are per-log conditions, so the old `break` was tearing down a *healthy* subscription. No busy-spin risk either — each iteration still awaits `sub.next()`.

### 🔗 Merge-order finding — please read before landing either PR

Three of the changes here are **fixes for hazards that #330 introduces**, not for hazards that exist on `main` today:

| #330 introduces | fixed by, in this PR |
|---|---|
| `reqwest` 0.13.2 → **0.13.4**, which attaches the request URL (containing the API key) to body errors | the `.text().await?` change below |
| `alloy-sol-types` 1.5.7 → **1.6.1**, which rejects surplus topics — and `main`'s `break` aborts the whole subscription on any decode error | `break` → `continue` |
| `sha3` 0.11 / `keccak` 0.2 arrive as new hash backends with nothing pinning expected digests | the new known-answer tests |

So this PR is only *nominally* independent of #330. **Landing #330 first lights up all three hazards.** Recommended order: **this PR first, then #330** — or both together.

This also resolves a puzzle in `ba658af`'s commit message (see the inline note): it describes the world *after* #330, which is why it doesn't match `main` today.

### ⚠️ CI's green `cargo-deny` here is stale

This PR's `cargo-deny (advisories)` ran before **[RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220)** (`ruint` — incorrect overflow flags and truncated shift amounts) was published. This branch carries `ruint 1.18.0` and would fail if re-run; #330 is what fixes it (`ruint 1.20.0`). Not a blocker for this PR, just don't read the green check as "advisories clean".

### 🧾 Pre-existing issues found next door

Not this PR's fault, and not blocking it — but the adversarial pass found these while tracing the changed code, and **several are more urgent than anything in this PR.** All line numbers are at this PR's head (`d568a3b`):

**Money correctness**
- [`etherscan_client/types.rs:95`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/etherscan_client/types.rs#L95) — `Decimal::new(self.value as i64, …)` casts a `u128` straight to `i64`. **Ten units of an 18-decimal token exceeds `i64::MAX` in base units and becomes a negative reconciliation amount.**
- [`polygon.rs:273`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/chain_client/polygon.rs#L273) — `u256_to_decimal` silently substitutes zero on failed conversion; a large or high-decimal payment records as **zero**.
- [`daemon/Cargo.toml:135`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/Cargo.toml#L135) — `[profile.*]` in a **non-root workspace member is silently ignored by Cargo**, so `overflow-checks = true` is *not* applied. Verified by reading the actual `rustc` invocation from `cargo build --release -vv`. ([Cargo Book: profiles are only honoured in the workspace root](https://doc.rust-lang.org/cargo/reference/profiles.html#profile-selection))

**Payment integrity**
- [`transfer_tracker.rs:179`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/chain/transfer_tracker.rs#L179) — database recording errors are logged and swallowed; the stream advances and is marked healthy. A transient SQLite lock makes an incoming payment **disappear** until manual reconciliation, possibly past invoice expiry.
- [`polygon.rs:892`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/chain_client/polygon.rs#L892) — `WS_MESSAGES_TIMEOUT_DURATION` treats *absence of matching Transfer logs* as websocket failure. A low-volume asset repeatedly tears down a healthy subscription and enters backoff, leaving unbackfilled gaps.
- [`polygon.rs:582`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/chain_client/polygon.rs#L582) — `U256::to::<u128>()` **panics** above `u128::MAX`, and the values come from unvalidated Pimlico bundler responses.

**Secret handling** (same theme as this PR — see the inline redaction note)
- [`configs/types.rs:429`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/configs/types.rs#L429) and [`etherscan_client/types.rs:37`](https://github.com/Kalapaja/Kalatori/blob/d568a3bc408dff0fc6cdb07640b65ce826d0613f/daemon/src/etherscan_client/types.rs#L37) — derived `Debug` over the plaintext key.

**Also worth knowing** (grounding pass, orthogonal to this PR): alloy transparently reconnects *and re-subscribes* (`service.rs` L66-107, 10 retries × 3 s by default), so a websocket blip is invisible to the daemon — **any `Transfer` emitted while disconnected is silently lost.** That's a missed-payment gap no code here can see.

---

*Review method: two blind passes, worktree-isolated, no tracked file modified. Findings marked "confirmed" were reproduced; nothing below is speculative unless labelled.*

daemon/src/etherscan_client.rs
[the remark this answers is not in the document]
foxcool: Done in `a34d886`. `.text().await?` now carries its own `inspect_err` logging `is_timeout` / `is_body` / `is_decode` — the same safe-flags-only rule as the `send()` arm, since reqwest 0.13.4 attaches the request URL to body errors too.

Also dropped `status` from the `send()` arm: those failures are `Kind::Request`, and `Error::status()` answers `Some` only for `Kind::Status`, so it was always `None`.

Dockerfile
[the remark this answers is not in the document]
foxcool: Agreed, and @kirushik raised the same thing. Filed as #342 — it wants a PR-only build with `push: false` rather than something this PR can carry.

Recorded in the verification table here as explicitly unverified, so it is not read as checked.

daemon/src/etherscan_client.rs
foxcool: Fixed in `fe91d91`, and you were right that the partial version was the worse outcome.

**Follow-up 1 — the other paths.** Both structs now redact. Worth noting the fix turned out to be *removing an inconsistency* rather than introducing a pattern: `BungeeApiConfig` and `ZeroExApiConfig` already declare `api_key: SecretString` (`configs/types.rs:568,574`) — Etherscan was the only key still on a bare `String`. So `EtherscanClientConfig.api_key` and the field on `EtherscanClient` are now `SecretString`, with a single `expose_secret()` at the query-building call site.

`GetAccountTokenTransactionsParams` keeps a plain `&str`, because `Serialize` is what puts the key in the query string — the derive is what had to go. Hand-written `Debug` prints `[REDACTED]` for that field and leaves the rest readable. This also retires the `// TODO: hide api_key field in logs` that was sitting on it.

**The sentinel test.** Two of them, one per struct, and I checked they actually bite: reverting the hand-written `Debug` to a derive fails with

```
api key leaked: GetAccountTokenTransactionsParams { module: "account", ..., api_key: "SENTINEL-ETHERSCAN-KEY-8f3a1c" }
```

**Follow-up 2 — `status` is dead.** Confirmed and removed, with the reason left in a comment so it does not come back. Read it off the pinned source: `status()` matches `Kind::Status` only, and a `send()` failure is `Kind::Request`.

daemon/src/etherscan_client.rs
foxcool: Rewritten in `a34d886`. You were right on the mechanism, and I read both crate sources to be sure rather than take the release notes' word for it:

- **0.13.2** — `text()`/`bytes()` map through `error::decode`, which is `Error::new(Kind::Decode, Some(e))`. No URL. So the `unwrap()` was a crash, not a leak.
- **0.13.4** — `do_bytes` ends `Err(err) => Err(crate::error::decode(err).with_url(*self.url))`. URL attached, key included.
- **`send()` at 0.13.2** — already carried it, which is why the `warn!` was the genuinely live defect.

The message now says all three, in that order, and names #330 as what turns the second one into a leak. Retitled too — "stop the API key reaching the logs, and stop the crash" — since the leak and the crash were separate defects with different severities, and the old title only described the one that was not yet real.

Dockerfile
foxcool: Both points taken; rewritten in `2d785b7`.

**The mechanism was stated backwards.** Verified your reading against the vendored sources: `aws-lc-sys` 0.40.0's `builder/` has zero occurrences of `USE_SYSTEM`, `OPENSSL_DIR`, `OPENSSL_LIB_DIR` or `pkg_config`. 0.43.0 has them. The comment now says what is true — inert at the locked version, pinning the intent ahead of ≥0.43.0, which #330 brings — and includes your point that even 0.43.0 only adopts a candidate carrying the `OPENSSL_IS_AWSLC` marker, so `libssl-dev` was never a candidate. The commit message says the same, since that is where the false claim actually lived.

**Second-order problem fixed too.** `AWS_LC_SYS_USE_SYSTEM = "0"` is now in `[env]` in the workspace `.cargo/config.toml` as well as the Dockerfile, so dev and CI builds cannot diverge from the image after the bump. Not `force`, so a deliberate override still wins.

The Dockerfile comment now points at `.cargo/config.toml` and vice versa, so neither reads as the whole policy.

daemon/src/chain_client/polygon.rs
foxcool: Added in `5999ae2` — provenance comment per constant, plus the standing rule.

Recorded exactly what you verified and against what: the userOp hash against a live `eth_call` to `getUserOpHash()` on EntryPoint v0.8 at `0x4337084D…F108` on Polygon mainnet with this fixture; the permit digest against USDC's on-chain `DOMAIN_SEPARATOR()` and `PERMIT_TYPEHASH()`; keccak256 against a from-scratch Keccak-f[1600] cross-checked with CPython `hashlib.sha3_256`. The v0.8 / `("ERC4337", "1")` pairing gets its own paragraph, with the reason it is worth pinning — a v0.6 construction on a v0.7+ EntryPoint passes its own test and fails in production.

Added one thing you did not ask for, because the provenance is only half the guarantee: a note that a failing value is re-derived the way its comment says, **never** refreshed from the new output. That is the move that would quietly turn these back into snapshots of a bug.

Comments only — no expected value changed.

daemon/src/chain_client/polygon.rs
foxcool: Agreed on all of it, and deliberately **not** doing it here. Filed as #341 with your failure scenario quoted.

Reasoning for the split: both suggestions — per-asset health rather than per-subscription, and a rate-limited `warn!` in place of `debug!` — are behaviour design that interacts with backfill (#333). With backfill a skipped event is a latency question; without it the event is gone. Tuning the visibility of a permanent loss before the mechanism that makes it not permanent means testing it twice, and the second test is the one that matters. It also needs per-asset state that does not exist in the stream loop today, which is more than a review-fix PR should grow.

So this PR keeps the strictly-better-than-`break` change and says so in the body, under "Deliberately not here", rather than leaving it looking finished.

One thing I did keep: the processing arm is `error!` with the transaction hash, since that log decoded — it is one of ours, and losing it is a payment nobody sees until expiry. Only the decode arm stays at `debug!`, on your reasoning that an undecodable log is shaped for someone else.

foxcool:
Revision pass done, @kirushik — all six threads addressed, branch force-pushed. Also took it out of draft.

**History changed**, so the old SHAs are gone. Only `ba658af` and `d568a3b` were rewritten (both because their *messages* stated something false); the other two original commits are untouched, and the new work is new commits.

| | |
|---|---|
| `a58aa6d` | unchanged |
| `4443d56` | unchanged |
| `a34d886` | was `ba658af` — message rewritten, plus the `status` removal and the body-error `inspect_err` |
| `2d785b7` | was `d568a3b` — message and comment rewritten, plus `.cargo/config.toml` |
| `fe91d91` | new — the rest of the key-redaction surface |
| `5999ae2` | new — KAT provenance |
| `c9148c2` | new — `ruint` 1.20.0, cherry-picked from #330 |

## Two things worth pulling out of the threads

**Your stale-green warning was load-bearing.** This branch could not be pushed at all without carrying the advisory fix — `ruint` 1.18.0 would have gone red on the first CI re-run and blocked a required check. `b553c3b2` is cherry-picked here so the branch can land first, which is the order you asked for. It resolves identically to #330; the only extra churn is six `ark-*` lockfile entries that never enter the build graph. `make cargo-deny` passes here now.

**The redaction fix was smaller than it looked**, because the codebase had already decided: `BungeeApiConfig` and `ZeroExApiConfig` declare `api_key: SecretString` and Etherscan was the sole holdout on a bare `String`. So `fe91d91` is convergence on an existing convention, not a new pattern — which is also why it is worth doing here rather than deferring.

## Filed rather than fixed

- **#341** — per-asset transfer health; your `continue`-is-silent finding, with the failure scenario quoted. Blocked behind #333 by design, reasoning in the thread.
- **#342** — no Docker build in PR CI. Both you and Copilot raised it.
- **#343** — `Decimal::new(self.value as i64, …)` at `etherscan_client/types.rs:118` from your "pre-existing next door" list. Writing it up moved it up my estimate of its severity: for an 18-decimal token the cast wraps at **ten units**, and this is the reconciliation path, so it records a payment wrong rather than displaying it wrong. The `#[expect(clippy::cast_possible_truncation)]` on the enclosing function is what has been keeping it quiet. Also noted `u256_to_decimal`'s zero-substitution as the same bug with the opposite symptom.

Not filed, because they are already tracked or owned elsewhere: MSRV (#331), WS reconnect gaps (#333), clean-clone lint/test (#339), the `[profile.*]` tables being ignored (#340), the second TLS stack (#335).

## Verification

`cargo-check`, `cargo-clippy` (`-D warnings`), `fmt --check`, `cargo-deny` all pass. Tests **212/212** — 210 before, two new redaction tests.

Both new tests were mutation-checked rather than assumed: reverting the hand-written `Debug` to a derive fails with the sentinel visible in the panic message.

Library claims were read from the vendored crate sources at the pinned versions, not from release notes — reqwest 0.13.2 vs 0.13.4 (`error.rs`, `async_impl/response.rs`) and aws-lc-sys 0.40.0 vs 0.43.0 (`builder/`). Every one of your three refuted claims checked out; details in the threads.

Merge order unchanged from your review: **this one, then #330** — or both together.

foxcool:
Rebased onto `main` after #345–#348 landed. Force-pushed; six commits, one fewer than before.

**Dropped the `ruint` cherry-pick.** #347 did the same bump on `main` — same version, same six `ark-*` 0.6.0 lockfile entries, same reasoning. My commit would have been a duplicate, and since the two were cut from different bases their patch-ids differ, so the rebase would have conflicted rather than dropping it silently. Removed before rebasing.

**One conflict git could not see.** The rebase reported no conflicts and GitHub reported `MERGEABLE`, but the result did not compile:

```
error[E0428]: the name `tests` is defined multiple times
  --> daemon/src/etherscan_client/types.rs:245:1
```

#348 added a `mod tests` to that file and so had this branch; both blocks merged cleanly as *text*. Merged the two modules by hand and folded the fix into `a340d533`, the commit that introduces my tests, so no commit in the series is individually broken.

Worth flagging as a class: a green `MERGEABLE` on a rebase across a rewritten file means the text merged, not that the result builds.

## The `break` → `continue` fix got more valuable, not less

`#348` rewrote `subscribe_transfers` around the `ConfirmationBuffer`, but both error arms still `break` on `main` (`polygon.rs:999` and `:1008`). With the buffer in place that is no longer only "tracking dies for every asset on this subscription" — the buffer is local to the `async_stream` block, so **`break` also discards every unreleased transfer in the trailing ~12 blocks**.

That is exactly the widened loss window described in the note added to #333. One undecodable log — an operator's non-standard ERC-20 emitting a surplus topic — now costs ~18 s of confirmed-but-unreleased payments, on every occurrence, with no replay cursor.

The `continue` landed on both arms correctly in the rewritten code; verified by reading the merged result, and the buffer is untouched on that path.

## #343 is fixed on `main`, closed

`into_incoming_transaction` now returns `Option` and rejects values past `Decimal`'s 96-bit mantissa and decimals past 28, with tests covering both plus the 18-decimal case above `i64::MAX`. Better than what the issue asked for — it draws the representability boundary explicitly instead of widening the cast. Nothing left for this PR.

## Verification after rebase

`cargo-check`, `cargo-clippy` (`-D warnings`), `fmt --check`, `cargo-deny` all pass. Tests **253/253** — the count moved because `main` gained tests from #345–#348.

Threads are unchanged and still answered; the rebase only moved their anchors.

kirushik (approved):
LGTM!

PR #335, opened 2026-07-31
build(deps): stop tracing-loki pulling in a second TLS stack

foxcool:
Manifest and lockfile only. No Rust changed.

## The problem

`tracing-loki`'s default features pull `reqwest/native-tls`, which on Linux means OpenSSL. So
a payment daemon whose every money path — RPC, Pimlico, Etherscan, merchant webhooks — runs on
rustls with aws-lc was also linking a second, C-implemented TLS library, for shipping logs.

## The change

Switching to the crate's `rustls` feature drops **eight packages and adds none**: `native-tls`,
`openssl`, `openssl-sys`, `openssl-macros`, `hyper-tls`, `tokio-native-tls`, `foreign-types`,
`foreign-types-shared`. Confirmed against the Linux target, where the OpenSSL chain actually
resolves — `cargo tree --target x86_64-unknown-linux-gnu -i openssl-sys` matches nothing.

No new crypto backend: `ring` was already in the tree via rustls through `alloy-transport-ws`,
checked before and after.

`compat-0-2-1` has to stay on. It is a default feature but not optional — tracing-loki
`compile_error!`s without it.

### Keeping the OS trust store — added after review

@kirushik caught what this quietly took away, and it was the right thing to block on.
tracing-loki's `rustls` feature maps to reqwest 0.12's `rustls-tls`, and **in that major
`rustls-tls` is hard-wired to `rustls-tls-webpki-roots`** — a bundled root bundle, not the
operating system's. native-tls read the OS store, so as originally written this swapped Loki's
certificate trust store as a side effect of dropping OpenSSL.

A self-hoster running Loki behind a corporate-CA certificate would have kept taking payments
while log pushes silently failed validation — losing observability exactly when they would want
it, with nothing in the release notes to explain why. It would also have made Loki the only
reqwest client in the daemon ignoring the system store: **the daemon's own reqwest is 0.13**,
whose `rustls` feature routes through `rustls-platform-verifier` and does read the OS store.

tracing-loki 0.2 exposes no native-roots variant — 0.2.6 and 0.2.7 have exactly `native-tls`
and `rustls`, nothing else — so the feature is enabled on reqwest 0.12 directly. Cargo unifies
features per resolved package, so naming it in this manifest applies to the instance
tracing-loki resolves to; the crate is never imported. Verified on the feature graph:

```
$ cargo tree --target x86_64-unknown-linux-gnu -e features -i reqwest:0.12.28
├── reqwest feature "__rustls"
│   ├── reqwest feature "rustls-tls-native-roots-no-provider"
│   │   └── reqwest feature "rustls-tls-native-roots" → kalatori
│   └── reqwest feature "rustls-tls-webpki-roots-no-provider"
│       └── reqwest feature "rustls-tls-webpki-roots" → tracing-loki feature "rustls"
```

Both sources. reqwest 0.12 adds them in independent branches, both defaulting on, so this is
**additive** — OS store plus bundled roots, never fewer roots than before this PR.

It costs nothing: `rustls-native-certs` was already in the tree via reqwest 0.13's platform
verifier, so the eight-removed / zero-added count survives the fix unchanged.

### Two follow-ups the review unlocked

- **`Dockerfile`**: `libssl-dev` is vestigial now that nothing links libssl. PR CI never builds
  the image (#342), so this was verified by building it locally rather than asserted. Worth
  noting while here: the runtime stage installs `ca-certificates` and nothing else, which is
  both why dropping `libssl-dev` is safe and why the native-roots fix above has a trust store
  to read in the shipped image — `rustls-native-certs` reads exactly what that package
  populates.
- **`deny.toml`**: the `OpenSSL` license allowance is gone, with a comment saying why it is
  absent. That turns a one-time removal into a standing guardrail against the stack returning
  through some future dependency's default features — which is exactly how it arrived.

## What this does not buy

`openssl 0.10.78` is affected by no open advisory
([RUSTSEC-2025-0022](https://rustsec.org/advisories/RUSTSEC-2025-0022.html) was patched in
≥ 0.10.72). The win is dropping runtime linkage to the C `libssl` — OS-level CVE surface
invisible to `cargo deny` — not closing a known hole. Recorded in the manifest so the next
reader gets the accurate version rather than inferring a stronger one.

## What this does not fix

`reqwest` 0.12 is still in the tree: tracing-loki requires `>=0.11.10, <0.13` while the daemon
and client are on 0.13. Two reqwest majors remain. What goes away is the second TLS
implementation and the C dependency.

## Not done

A private-CA HTTPS regression test. tracing-loki exposes no handle on its reqwest client, so
end-to-end coverage means generating a CA, standing up an HTTPS server, pointing
`SSL_CERT_FILE` at the bundle and asserting a push lands — two dev-dependencies and a TLS test
server for one assertion. The invariant that actually broke here is a feature-graph property,
checked above. Happy to add the runtime test as a follow-up if it is wanted.

## Verification

Rebased onto current `main` (0.9.4).

| Check | Result |
|---|---|
| `cargo check --workspace --all-targets` | pass |
| clippy, `RUSTFLAGS="-Dwarnings"` | pass |
| `cargo deny check bans licenses sources` | ok |
| tests | 292, unchanged — no Rust touched |
| `docker build --platform linux/amd64` | pass |

The claim this PR actually makes is about linkage, so it is checked on the shipped artifact
rather than inferred from the lockfile:

```
$ ldd /app/kalatori
        libsqlite3.so => /usr/local/lib/libsqlite3.so
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6
        /lib64/ld-linux-x86-64.so.2
```

No `libssl`, no `libcrypto`. Note that `libssl.so.3` and `libcrypto.so.3` do still exist inside
the runtime image — they come with `debian:bookworm-slim` and its `ca-certificates`, not from
anything we build. The point is that the daemon no longer depends on them, which is what takes
the OS-level CVE surface off our runtime.


daemon/Cargo.toml
```diff
@@ -89,7 +89,14 @@ parity-scale-codec = { version = "3", features = ["derive"] }
...
 tracing-subscriber = { version = "0.3", features = ["env-filter", "time", "json"] }
-tracing-loki = "0.2"
+# Default features pull reqwest/native-tls, i.e. OpenSSL on Linux, alongside the
+# rustls/aws-lc stack every money path already uses. Log shipping does not get to
+# add a second TLS implementation. `compat-0-2-1` is not optional: the crate
+# refuses to compile without it.
+tracing-loki = { version = "0.2", default-features = false, features = [
+    "compat-0-2-1",
+    "rustls",
+] }
```
kirushik, line 108: **Selecting `rustls` here also silently swaps Loki's certificate trust store from the operating system's to a bundled one. Both review passes found this independently, and it's the one thing holding the PR back.**

The chain:

```
tracing-loki 0.2.6  "rustls"
  └── reqwest 0.12  "rustls-tls"
        └── rustls-tls = ["rustls-tls-webpki-roots"]   ← hard-wired in reqwest 0.12
              └── roots come from the bundled webpki-roots crate, NOT the OS store
```

That last step is the surprise: in reqwest 0.12 the plain `rustls-tls` feature *is* the webpki-roots variant ([v0.12.28 `Cargo.toml`](https://github.com/seanmonstar/reqwest/blob/v0.12.28/Cargo.toml)), and [`client.rs` L692-694](https://github.com/seanmonstar/reqwest/blob/v0.12.28/src/async_impl/client.rs) extends the store from `webpki_roots::TLS_SERVER_ROOTS` only. Previously native-tls/OpenSSL read the **OS** trust store.

**Concrete failure:**

> A self-hoster runs Loki at `https://loki.internal`, with a certificate chaining to a corporate CA installed in the container's trust store. They upgrade Kalatori. Payments keep flowing normally. Loki pushes start failing certificate validation, queued logs are dropped, and **nothing in the payment path signals a problem** — they lose observability precisely when they'd want it.

No configuration change on their side triggers this, and nothing in the release notes would tell them. That's what makes it worth blocking on: it's not a bug, it's an unannounced contract change riding along with a dependency cleanup.

There's a smaller consistency cost too — Loki becomes the only reqwest client in the daemon that ignores the system store, so "does Kalatori trust my private CA?" gets a per-subsystem answer.

**Suggested fix** (small):

```toml
tracing-loki = { version = "0.2", default-features = false, features = [
    "compat-0-2-1",
    "rustls-native-roots",   # or whichever variant tracing-loki exposes that maps
                             # to reqwest's rustls-tls-native-roots
] }
```

If tracing-loki doesn't expose a native-roots variant at 0.2.6, the alternatives are enabling `reqwest/rustls-tls-native-roots` directly (feature unification will apply it) or supporting an explicit operator-supplied CA bundle. Either way, please **say so in the PR description** — the OpenSSL removal is a clear win and shouldn't be entangled with an unmentioned trust-store change.

Worth adding a private-CA HTTPS regression test at the same time; there's currently nothing exercising Loki transport at all.

One nice thing this comment block gets right, by the way: noting that `compat-0-2-1` isn't optional. It genuinely isn't — the crate `compile_error!`s without it — and that's exactly the kind of non-obvious constraint that deserves the inline note.

daemon/Cargo.toml
```diff
@@ -89,7 +89,14 @@ parity-scale-codec = { version = "3", features = ["derive"] }
 # Tracing
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "time", "json"] }
-tracing-loki = "0.2"
+# Default features pull reqwest/native-tls, i.e. OpenSSL on Linux, alongside the
+# rustls/aws-lc stack every money path already uses. Log shipping does not get to
+# add a second TLS implementation. `compat-0-2-1` is not optional: the crate
+# refuses to compile without it.
```
kirushik, line 97: **Every claim in this comment checks out — including the ones I expected to find overstated.** Recording the verification here so it doesn't have to be redone:

| Claim | Result |
|---|---|
| default features pull `reqwest/native-tls` | ✅ confirmed at tracing-loki 0.2.6 |
| that means OpenSSL on Linux | ✅ confirmed |
| the money paths already run rustls/aws-lc | ✅ confirmed — and aws-lc is installed process-wide at `main.rs:197`, before logger init |
| exactly 8 packages removed, 0 added | ✅ confirmed from the lockfile diff |
| `openssl-sys` gone from the Linux target graph | ✅ `cargo tree -e features -i openssl-sys` matches nothing |
| `compat-0-2-1` is mandatory | ✅ the crate `compile_error!`s without it |

Two things the comment could usefully add, since it's already the place a future reader will look:

1. **What this does *not* buy.** `openssl 0.10.78` is affected by no open advisory — [RUSTSEC-2025-0022](https://rustsec.org/advisories/RUSTSEC-2025-0022.html) was patched in ≥ 0.10.72. The real win is dropping runtime linkage to the C `libssl` (OS-level CVE surface, invisible to `cargo deny`), not closing a known hole. The PR doesn't overclaim — but stating the actual benefit makes the change harder to argue with later.

2. **The trust-store consequence**, per the other comment. If someone reverts this in six months because their private-CA Loki broke, this comment is where they'll look for the reasoning.

Two cheap follow-ups now unlocked, worth doing in this PR while the context is loaded: [`Dockerfile:11`](https://github.com/Kalapaja/Kalatori/blob/b64775fad8d4f6f27d5d41ec98ea2eca53a113af/Dockerfile#L11)'s `libssl-dev` is vestigial, and dropping the `"OpenSSL"` allowance at [`deny.toml:29`](https://github.com/Kalapaja/Kalatori/blob/b64775fad8d4f6f27d5d41ec98ea2eca53a113af/deny.toml#L29) turns `cargo deny` into a standing guardrail against the stack coming back — which is a stronger guarantee than the one-time removal this PR is currently claiming.

Minor: upstream tracing-loki's CI only builds default features ([`build.yaml`](https://github.com/hrxi/tracing-loki/blob/master/.github/workflows/build.yaml)), so the `rustls` path is untested upstream. In practice the risk is near zero — `grep 'cfg(feature'` over tracing-loki's `src/` returns nothing except the `compat-0-2-1` `compile_error!`, so there's no feature-gated code path to get wrong — but it's worth knowing you're the one testing it.

kirushik (commented):
## Review summary

Two independent passes, deliberately not shown each other's output: an **adversarial code pass** (GPT-5.6 `sol` via codex, high reasoning effort, working from the resolved feature graph and the pinned crate sources on disk) and an **upstream-grounding pass** (Claude Opus with web access, reading reqwest's and tracing-loki's actual `Cargo.toml` at the pinned tags).

**Both landed on the same blocking finding, from different directions** — that convergence is the strongest signal in this review batch.

**Verdict: the change does exactly what it says, and carries one behaviour change it doesn't mention.** The fix is a one-line feature swap.

### ✅ Every factual claim in the description checks out

I want to be explicit about this, because the description makes several specific numeric claims and they all survived checking:

- **Exactly 8 packages removed, 0 added** — confirmed from the lockfile diff.
- **`cargo tree -e features -i openssl-sys` matches nothing** on the Linux target. OpenSSL and native-tls are genuinely gone, not merely deprioritised.
- **No dependency re-enables `native-tls` through feature unification** — the failure mode where a change like this silently accomplishes nothing. Checked; it doesn't happen here.
- **`ring` genuinely pre-existed**; this PR didn't introduce it.
- **`compat-0-2-1` really is mandatory** — tracing-loki emits a `compile_error!` without it, so keeping it isn't cargo-culting.
- The lockfile is properly **regenerated, not hand-edited** (`cargo metadata --locked` agrees).

### ✅ Two failure modes I went looking for, and did *not* find

Recording these so nobody has to re-check them:

1. **reqwest is not left without a TLS backend.** That's the nasty one — it fails at *runtime* on `https://`, not at compile time, so CI stays green while log shipping dies. Not the case here: reqwest 0.12.28 retains an active rustls backend.
2. **No `CryptoProvider` conflict.** reqwest 0.12.28 prefers `CryptoProvider::get_default()` before falling back to ring — *"Allow user to have installed a runtime default. If not, we use ring"* ([`client.rs` L762-771](https://github.com/seanmonstar/reqwest/blob/v0.12.28/src/async_impl/client.rs)) — and `main.rs:197` installs `aws_lc_rs::default_provider()` **before** `logger::initialize` at `main.rs:207`. So Loki runs on aws-lc-rs and the linked `ring` is dead code.

   ⚠️ **That ordering is load-bearing and undocumented.** Swapping those two lines in `main.rs` would silently move Loki onto a different crypto backend. Worth a comment at the call site.

### ⚠️ The blocking issue

Loki silently stops consulting the system CA store — see the inline comment. Both passes found it independently.

### 🧾 Smaller follow-ups

- **[`Dockerfile:11`](https://github.com/Kalapaja/Kalatori/blob/b64775fad8d4f6f27d5d41ec98ea2eca53a113af/Dockerfile#L11) `libssl-dev` is now vestigial** — nothing in the resolved Linux graph consumes OpenSSL. Removing it makes the guarantee visible in the image definition.
- **[`deny.toml:29`](https://github.com/Kalapaja/Kalatori/blob/b64775fad8d4f6f27d5d41ec98ea2eca53a113af/deny.toml#L29) still allows the `"OpenSSL"` license.** Dropping that entry converts `cargo deny` into a *guardrail against reintroducing the stack* — which is a nicer outcome than this PR currently claims for itself.
- **No test or CI check exercises Loki delivery at all.** No unit test, example or workflow enables `loki_url`, receives a push, or exercises HTTPS transport. So a TLS or trust-root regression passes `check`, `clippy`, unit tests and integration tests, and then disables log shipping in production. Given this PR *is* a TLS change to Loki, that's the gap that makes the CA regression above undetectable. Worth its own issue rather than blocking here — a loopback HTTPS receiver asserting a record arrives at `/loki/api/v1/push` would cover it.
- **`docs/architecture.md` isn't updated.** The repo's documentation policy is binding and this changes TLS posture; right now the only description of the new arrangement is an inline `Cargo.toml` comment.

### 📌 Pre-existing context

The split trust-root story predates this PR: Polygon WSS already uses webpki-roots via `tokio-tungstenite 0.28`, while Asset Hub RPC and the reqwest 0.13 clients use the system store. **This PR adds a third divergent client rather than creating the divergence** — but it's a good moment to decide what the daemon's TLS policy actually is and write it down once.

### 🔎 Verification caveat, stated honestly

Neither pass could compile the daemon: `metadata.scale` and `static/index.html` are gitignored generated inputs absent from a clean checkout, so `cargo check`/`clippy` exit 101 before reaching daemon code. The TLS conclusions come from the resolved feature graph and the pinned crate sources, which is solid evidence for *these* questions — but **nobody has built this branch locally.** CI is green here apart from the `cargo-deny` advisories failure inherited from `main` ([RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220) — `ruint`, fixed by #330, unrelated to this change).

daemon/Cargo.toml
foxcool: Fixed, and you were right that this was the thing to block on. I traced the chain you gave and it holds exactly as described — reqwest 0.12.28's `rustls-tls = ["rustls-tls-webpki-roots"]`, and `client.rs` extending the store from `webpki_roots::TLS_SERVER_ROOTS` only.

One thing your comment left implicit that makes the case stronger: **the daemon's own reqwest is 0.13**, where `rustls = ["__rustls-aws-lc-rs", "dep:rustls-platform-verifier", "__rustls"]` — the platform verifier, which does read the OS store. So this was not "everything moves to bundled roots"; it was Loki alone diverging from every other client in the process. Your per-subsystem-answer point, but sharper.

**tracing-loki 0.2 exposes no native-roots variant** — 0.2.6 and 0.2.7 both have exactly `native-tls = ["reqwest/native-tls"]` and `rustls = ["reqwest/rustls-tls"]`, nothing else. So of the alternatives you listed, it had to be enabling the reqwest feature directly:

```toml
loki-transport-roots = { package = "reqwest", version = "0.12", default-features = false, features = [
    "rustls-tls-native-roots",
] }
```

Cargo unifies features per resolved package, so this applies to the instance tracing-loki resolves to; the crate is never imported. Verified on the feature graph rather than assumed:

```
$ cargo tree --target x86_64-unknown-linux-gnu -e features -i reqwest:0.12.28
reqwest v0.12.28
├── reqwest feature "__rustls"
│   ├── reqwest feature "rustls-tls-native-roots-no-provider"
│   │   └── reqwest feature "rustls-tls-native-roots"
│   │       └── kalatori v0.9.4
│   └── reqwest feature "rustls-tls-webpki-roots-no-provider"
│       └── reqwest feature "rustls-tls-webpki-roots"
│           └── reqwest feature "rustls-tls"
│               └── tracing-loki feature "rustls"
```

Both root sources, which is the point: reqwest 0.12 adds them in two independent `if` blocks (`tls_built_in_certs_webpki` / `tls_built_in_certs_native`, both defaulting to `true`), so this is additive — OS store **plus** bundled roots, never fewer roots than before this PR. Your self-hoster keeps working.

**It costs nothing.** `rustls-native-certs` was already in the tree via reqwest 0.13's platform verifier, so the lockfile delta is unchanged: still eight packages removed, none added.

Both follow-ups from your other comment are done here: `libssl-dev` dropped from the Dockerfile (verified by building the image locally, since PR CI never builds it — #342), and the `OpenSSL` allowance dropped from `deny.toml`, which as you say is the stronger guarantee. `cargo deny check licenses` passes.

**Not done: the private-CA regression test.** I'd rather say so than quietly skip it. tracing-loki gives no handle on its reqwest client, so testing this end-to-end means generating a CA, standing up an HTTPS server, pointing `SSL_CERT_FILE` at the bundle and asserting a log push lands — two new dev-dependencies and a TLS test server for one assertion. Happy to do it as a follow-up if you think it earns its keep; my read is that the feature-graph check above is the invariant that actually broke, and it is not something a runtime test would have caught any earlier.

daemon/Cargo.toml
foxcool: Both additions made, and both follow-ups done in this PR as you suggested.

**What this does not buy** is now in the manifest comment: `openssl 0.10.78` is affected by no open advisory — RUSTSEC-2025-0022 was patched in ≥ 0.10.72 — and the real win is dropping runtime linkage to the C `libssl`, an OS-level CVE surface `cargo deny` cannot see. You're right that stating the actual benefit makes the change harder to argue with later than letting a reader infer a stronger one.

**The trust-store consequence** is in the manifest too, at length, per the other thread. That is exactly where someone reverting this in six months will look.

**`Dockerfile:11`** — `libssl-dev` dropped. Since PR CI never builds the image (#342), I built it locally rather than assert it: `docker build --platform linux/amd64` gets through `apt-get install` and on into the Rust build with no OpenSSL headers present.

**`deny.toml:29`** — `OpenSSL` allowance dropped, with a comment saying why it is absent rather than leaving a silent gap for someone to "fix" by re-adding it. `cargo deny check bans licenses sources` → `ok`. Agreed this is the stronger guarantee: the one-time removal becomes a standing guardrail against the stack returning through a future dependency's default features, which is precisely how it arrived.

Thanks for recording the verification table — it saved re-deriving all six, and the two that mattered most (`compat-0-2-1` being mandatory, and the eight-removed/zero-added count) both still hold after the native-roots change. `rustls-native-certs` was already in the tree via reqwest 0.13, so honouring the OS store added no packages.

Noted on upstream CI only building default features. Worth knowing, and your reasoning for why the risk is near zero — no feature-gated code paths in `src/` beyond the `compat-0-2-1` `compile_error!` — matches what I see. Recorded here rather than in the manifest, since it is a fact about tracing-loki's CI rather than about our configuration.

kirushik (commented):
Ok, so I've merged #330, so this one needs some manual work on conflicts.
Also, please note that docs/architecture.md still hadn't been updated, and the *load-bearing* "CryptoProvider before logger init" ordering in main.rs still has no comment.

Needs a merge and a small-ish fixup commit

foxcool:
Merged and fixed up — three commits: the merge, then the comment and the doc.

**Conflicts.** Only `Cargo.lock`, and I regenerated it from main's rather than hand-merging — a lockfile resolved by hand is a lockfile nobody can reproduce. Since #330 moved alloy to 2.x underneath this branch, I re-verified the claims on the merged tree instead of assuming they survived:

- reqwest 0.12.28 still resolves with **both** `rustls-tls-native-roots-no-provider` and `rustls-tls-webpki-roots-no-provider` — the OS trust store is still honoured for log shipping;
- `cargo tree --target x86_64-unknown-linux-gnu -i openssl-sys` still matches nothing;
- `cargo deny check bans licenses sources` passes with the OpenSSL allowance still gone;
- the delta over main is unchanged: same eight packages removed, none added.

**The ordering comment.** You called it load-bearing and that is exactly right — and this PR is what makes it load-bearing, which is the part that deserved writing down. rustls takes one process-wide default provider, first install wins. Before this PR, log shipping was on native-tls and never touched rustls. After it, `tracing-loki` resolves reqwest 0.12, whose rustls feature set pulls **ring** — while the daemon installs **aws-lc**. Confirmed on the merged tree: reqwest 0.13.4 resolves `__rustls-aws-lc-rs`, reqwest 0.12.28 resolves `__rustls-ring`.

So initialising the logger before that call would let it install ring, our `install_default()` would fail, and the `.unwrap()` would panic — which under the panic gate means the daemon refuses to start. The comment says that, and says explicitly that the neighbouring `#[expect]` reason ("cannot already be set") is only true *because* of the ordering.

**`docs/architecture.md`** gains a TLS section, per the documentation policy. Beyond the single-stack decision and the ordering, it records the thing a future reader will actually trip on: two reqwest majors are in the tree and they reach the OS trust store by different routes.

| Consumer | reqwest | How roots are chosen |
|---|---|---|
| Money paths | 0.13 | `rustls` → `rustls-platform-verifier` → OS store |
| Log shipping | 0.12 | `rustls-tls` → bundled webpki, **plus** `rustls-tls-native-roots` enabled explicitly → OS store too |

Stated that way because the failure mode is deletion: someone tidying up an apparently unused `reqwest` entry in `daemon/Cargo.toml` would silently put log shipping on a different trust store from every money path, and nothing would fail.

Clippy under `-Dwarnings`, fmt, 301 tests, `cargo deny` — all green on the merged tree.

PR #336, opened 2026-07-31
fix(configs): say so when falling back to the public default endpoints

foxcool:
Two `tracing::warn!` calls plus seven tests. Grew out of the #333 investigation rather than the
#330 review, but belongs with this batch of small reliability fixes.

## The problem

`set_default_chains_if_missing()` substitutes `DEFAULT_POLKADOT_ASSET_HUB_ENDPOINTS` and
`DEFAULT_POLYGON_ENDPOINTS` without logging anything at all — no WARN, no INFO. An operator
who never configured endpoints has no way to learn which node their payments depend on, short
of reading the source.

This is not hypothetical. The instance in #333 spent **three months** on
`wss://polygon-bor-rpc.publicnode.com`, and nothing in its logs said so. That instance was
observed reconnecting on a roughly 60-second period — one observation on one instance, not a
measured property of the provider, and nothing here depends on the cause. The subscription
events lost to those reconnects were the actual bug; the reason nobody looked is that the
endpoint choice was invisible.

## The change

Warns once per chain, naming the chain and the endpoints it fell back to, and saying plainly
that these are free public defaults not meant for production.

Checked before writing it that the line actually lands: `logger::initialize` runs before
`chains_config_with_prefix` in `main.rs`. Had the order been the other way round, the fix
would have been useless and silently so.

**Swaps too, after review.** `ZeroExApiConfig::default()` hands out the same free provider
over `https` instead of `wss`, and startup *always* constructs `SwapsClients` — so an operator
who configured both payment chains properly still read a clean log while depending on a public
node for swaps. Partial coverage on a visibility feature is worse than none, because a clean
log then reads as "I am fully configured". The URL is now `DEFAULT_ZERO_EX_RPC_URL` in
`consts.rs`, beside the `wss://` entry it shares a host with, and `swaps_config_with_prefix`
announces it at the same point in startup.

Deliberately a warning and not the config rework the `TODO` above that `Default` impl asks
for: making the 0x client optional or its RPC required is a startup-behaviour change and does
not belong in a visibility fix.

## Tests

Seven, all asserting on captured output via `tracing_test`, and all going through
`logs_assert` rather than substring searches over the buffer — matching level, message and the
`chain` field **together**, so an ambiguous INFO satisfies nothing and a warning fired for the
wrong chain fails on identity rather than passing on a substring.

- every unconfigured chain announced **exactly once**;
- Polygon configured → only Asset Hub announced, and the **reverse**, since asymmetric bugs
  pass one direction and fail the other;
- `endpoints: []` — built by deserializing JSON, so it travels the explicit-empty-sequence
  path rather than `#[serde(default)]`;
- the **loader** entry point, not just the method, with a non-existent config dir and a prefix
  unique to the test;
- both swaps branches, defaulted and configured — the negative one through the same
  level-aware helper as the positive, not a bare `logs_contain`;
- the **swaps loader**, `swaps_config_with_prefix`. Deleting its call to
  `warn_if_zero_ex_rpc_is_public_default` previously left every test green, which is this
  PR's own bug reintroduced through the one line nothing looked at.

Verified by mutation, both directions: dropping the level to `info!` fails five of seven;
reporting a fixed chain instead of the one that fell back fails four, on the count assertion
and the identity assertion respectively.

Note: `daemon/src/configs/types.rs` gained a test module on `main` since this branch opened,
so these live in it rather than a second one.

## Adjacent question, not decided here

@kirushik raised in #333 whether free public providers belong in compiled-in defaults at all.
That is a product decision and this PR does not take it — it only makes the current behaviour
visible. If the answer is "they should not", these warnings become a hard error at startup and
the change is a small one.

Also not covered, and I don't think it belongs here: `ZeroExApiConfig`'s `api_key` defaults to
`""`, so an unconfigured 0x client is not merely degraded but broken at its first
authenticated call. That is the `TODO`'s territory, not a visibility gap.

## Verification

Rebased onto current `main` (0.9.4).

| Check | Result |
|---|---|
| `cargo check --workspace --all-targets` | pass |
| clippy, `RUSTFLAGS="-Dwarnings"`, all targets and features | pass |
| tests | **299** (292 before, 7 new) |
| CI | 8/8 green |

### DIFF
--- a/daemon/src/configs/types.rs
+++ b/daemon/src/configs/types.rs
@@ -264,6 +264,16 @@ impl ChainsConfig {
                     ChainType::Polygon => DEFAULT_POLYGON_ENDPOINTS,
                 };
 
+                // These are free public endpoints with no availability guarantee.
+                // An operator who never configured any has no other way to find
+                // out which node their payments depend on.
+                tracing::warn!(
+                    chain = %chain,
+                    endpoints = ?endpoints,
+                    "No endpoints configured, falling back to free public defaults. \
+                     Configure your own endpoints for production use."
+                );
+
                 chain_config.endpoints = endpoints
                     .iter()
                     .map(|s| ChainEndpoint::Universal(s.to_string()))
@@ -823,4 +833,72 @@ pub struct SwapsConfig {
     pub zero_ex: ZeroExApiConfig,
     #[serde(default)]
     pub fees: Option<IntegratorFees>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn empty_chains_config() -> ChainsConfig {
+        ChainsConfig {
+            chains: HashMap::new(),
+        }
+    }
+
+    #[test]
+    #[tracing_test::traced_test]
+    fn default_endpoints_are_announced_when_none_configured() {
+        let mut config = empty_chains_config();
+
+        config.set_default_chains_if_missing();
+
+        for chain in ChainType::iter() {
+            assert!(
+                config.chains[&chain]
+                    .get_random_requests_endpoint()
+                    .is_some(),
+                "{chain} was left without endpoints"
+            );
+        }
+
+        // The point of the warning is that an operator who configured nothing
+        // can still tell from the log which node their payments depend on.
+        assert!(logs_contain(
+            "falling back to free public defaults"
+        ));
+        assert!(logs_contain(
+            "wss://polygon-bor-rpc.publicnode.com"
+        ));
+        assert!(logs_contain(
+            "wss://asset-hub-polkadot-rpc.n.dwellir.com"
+        ));
+    }
+
+    #[test]
+    #[tracing_test::traced_test]
+    fn configured_endpoints_are_left_alone_and_unannounced() {
+        let configured = "wss://polygon.example.internal";
+        let mut config = empty_chains_config();
+        config.chains.insert(
+            ChainType::Polygon,
+            ChainConfig {
+                endpoints: vec![ChainEndpoint::Universal(configured.to_string())],
+                ..ChainConfig::default()
+            },
+        );
+
+        config.set_default_chains_if_missing();
+
+        assert_eq!(
+            config.chains[&ChainType::Polygon].get_random_requests_endpoint(),
+            Some(configured.to_string()),
+        );
+        assert!(!logs_contain(
+            "wss://polygon-bor-rpc.publicnode.com"
+        ));
+        // Asset Hub was still unconfigured, so it does get the warning.
+        assert!(logs_contain(
+            "wss://asset-hub-polkadot-rpc.n.dwellir.com"
+        ));
+    }
 }
--- a/daemon/src/configs/types.rs
+++ b/daemon/src/configs/types.rs
@@ -20,6 +20,10 @@ use crate::types::{
     ChainType,
     DetectedShopPlatform,
 };
+use crate::utils::logging::{
+    category,
+    operation,
+};
 
 use super::consts::{
     DEFAULT_ALLOW_INSECURE_ENDPOINTS,
@@ -268,7 +272,21 @@ impl ChainsConfig {
                 // These are free public endpoints with no availability guarantee.
                 // An operator who never configured any has no other way to find
                 // out which node their payments depend on.
+                //
+                // WARN rather than ERROR is a deliberate choice: this is the
+                // documented dev-mode default and the daemon runs correctly on
+                // it, so it is degraded state rather than a critical failure —
+                // the WARN row of the level table in docs/conventions.md. An
+                // operator who wants it to be fatal can promote it downstream
+                // on `error.category = "config"`.
+                //
+                // `endpoints` is the hardcoded default slice, never the
+                // operator's own value: configured endpoint URLs can carry
+                // credentials in userinfo or the path, so only compiled-in
+                // constants are ever printed here. Tests pin this down.
                 tracing::warn!(
+                    error.category = category::CONFIG,
+                    error.operation = operation::LOAD_CHAIN_ENDPOINTS,
                     chain = %chain,
                     endpoints = ?endpoints,
                     "No endpoints configured, falling back to free public defaults. \
@@ -840,6 +858,68 @@ mod tests {
         );
         logs_assert(|logs| assert_swaps_warnings(logs, 1));
     }
+
+    /// A URL an operator configured is theirs, and can carry credentials in
+    /// userinfo or in the path. The rule these warnings follow is that only
+    /// compiled-in constants are ever printed — this pair is what holds the
+    /// rule in place, since every other test here asserts on what *is* logged
+    /// and would stay green while a secret leaked alongside it.
+    ///
+    /// Only URL-borne secrets need a runtime check. `ZeroExApiConfig::api_key`
+    /// is a `SecretBox<str>`, so logging it is a compile error and its `Debug`
+    /// is redacted — a test for it could never fail. `rpc_url` beside it is a
+    /// plain `String`, which is why `?self.zero_ex` would still leak and why
+    /// the swaps case below is worth pinning.
+    const CREDENTIAL_SENTINELS: &[&str] = &["hunter2", "s3cr3t-path"];
+
+    fn assert_no_sentinel_leaked(logs: &[&str]) -> Result<(), String> {
+        for sentinel in CREDENTIAL_SENTINELS {
+            if logs
+                .iter()
+                .any(|log| log.contains(sentinel))
+            {
+                return Err(format!(
+                    "configured value `{sentinel}` reached the log"
+                ));
+            }
+        }
+
+        Ok(())
+    }
+
+    #[test]
+    #[tracing_test::traced_test]
+    fn a_configured_chain_endpoint_never_reaches_the_log() {
+        let mut config = chains_config_from_json(r#"{"chains":{}}"#);
+        config.chains.insert(
+            ChainType::Polygon,
+            configured("wss://operator:hunter2@polygon.example.internal/s3cr3t-path"),
+        );
+
+        config.set_default_chains_if_missing();
+
+        logs_assert(|logs| {
+            // Asset Hub still fell back, so the warning path really did run.
+            assert_warned_once(logs, ChainType::PolkadotAssetHub)?;
+            assert_no_sentinel_leaked(logs)
+        });
+    }
+
+    #[test]
+    #[tracing_test::traced_test]
+    fn a_configured_swaps_rpc_never_reaches_the_log() {
+        let config = SwapsConfig {
+            zero_ex: ZeroExApiConfig {
+                api_key: "not-a-real-key".into(),
+                rpc_url: "https://operator:hunter2@rpc.example.internal/s3cr3t-path".to_string(),
+            },
+            ..SwapsConfig::default()
+        };
+
+        config.warn_if_zero_ex_rpc_is_public_default();
+
+        logs_assert(assert_no_sentinel_leaked);
+    }
 }
 
 fn default_host() -> IpAddr {
@@ -1095,8 +1175,18 @@ impl SwapsConfig {
     /// warning exists to abolish.
     pub(super) fn warn_if_zero_ex_rpc_is_public_default(&self) {
         if self.zero_ex.rpc_url == DEFAULT_ZERO_EX_RPC_URL {
+            // Logs the constant, not `self.zero_ex.rpc_url`. The guard makes
+            // the two equal today, so this reads as a pointless substitution —
+            // it is not. Printing the field would mean the leak is prevented
+            // only by the guard's exact-equality, and any later loosening of
+            // that guard (to a host match, say) would start printing operator
+            // URLs, which can carry credentials in userinfo or the path.
+            // Sourcing the value from the constant makes that impossible by
+            // construction rather than by careful reading.
             tracing::warn!(
-                rpc_url = %self.zero_ex.rpc_url,
+                error.category = category::CONFIG,
+                error.operation = operation::LOAD_SWAPS_RPC,
+                rpc_url = DEFAULT_ZERO_EX_RPC_URL,
                 "Swaps are using a free public RPC endpoint. \
                  Set swaps.zero_ex.rpc_url to your own endpoint for production use."
             );
--- a/daemon/src/configs/types.rs
+++ b/daemon/src/configs/types.rs
@@ -767,6 +767,41 @@ mod tests {
         });
     }
 
+    /// Counterpart of `fallback_warnings_for` for the swaps branch. The
+    /// negative case goes through this too: `!logs_contain(..)` would also be
+    /// satisfied by the warning being emitted at the wrong level, which is the
+    /// failure this pair is supposed to catch.
+    fn swaps_fallback_warnings(logs: &[&str]) -> usize {
+        logs.iter()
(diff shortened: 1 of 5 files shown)


daemon/src/configs/types.rs
```diff
@@ -184,6 +184,16 @@ impl ChainsConfig {
...
                 };
 
+                // These are free public endpoints with no availability guarantee.
+                // An operator who never configured any has no other way to find
+                // out which node their payments depend on.
+                tracing::warn!(
+                    chain = %chain,
+                    endpoints = ?endpoints,
+                    "No endpoints configured, falling back to free public defaults. \
+                     Configure your own endpoints for production use."
```
kirushik, line 293: **This warning covers the chain endpoints but not the sibling fallback 400 lines down — and that one uses the same free provider. Both review passes flagged it independently.**

[`configs/types.rs:595`](https://github.com/Kalapaja/Kalatori/blob/a82d0dbb8fa1101e7a91cfd192e8260d2c64acb3/daemon/src/configs/types.rs#L595):

```rust
impl Default for ZeroExApiConfig {
    fn default() -> Self {
        Self {
            api_key: "".into(),
            rpc_url: "https://polygon-bor-rpc.publicnode.com".to_string(),
        }
    }
}
```

Same host as [`consts.rs:22`](https://github.com/Kalapaja/Kalatori/blob/a82d0dbb8fa1101e7a91cfd192e8260d2c64acb3/daemon/src/configs/consts.rs#L22)'s `wss://polygon-bor-rpc.publicnode.com`, over https instead of wss. Startup **always** constructs `SwapsClients`, so omitting `swaps.zero_ex` puts the daemon on an unconfigured public RPC with no warning at all.

**Concrete failure:**

> An operator reads the new warning, configures endpoints for both payment chains properly, restarts, sees a clean log — and is still depending on a free public RPC for swaps, with no way to discover it short of reading the source. Which is the exact condition this PR exists to abolish.

That's why I'd call this a half-fix rather than a complete one: partial coverage on a visibility feature is worse than none, because a *clean* log now reads as "I'm fully configured."

**Suggested:** emit the same warning at the 0x client's post-merge defaulting boundary, or make the 0x RPC explicit/required, or make the client optional so it isn't constructed when unconfigured. Whichever you pick, keep the rule this warning already gets right: **log only the hardcoded default, never a configured value** — operator-supplied endpoint URLs can carry API keys in the path or userinfo, so printing them back would turn a visibility fix into a credential leak.

Two smaller things on this block:

1. **Structured fields.** Per [`docs/conventions.md`](https://github.com/Kalapaja/Kalatori/blob/a82d0dbb8fa1101e7a91cfd192e8260d2c64acb3/docs/conventions.md) this should carry the constant-backed `error.category` / `error.operation` fields from `daemon/src/utils/logging.rs` alongside the ad-hoc `chain` / `endpoints`. Without them, production alert routing can't classify "started in a degraded configuration" — and an alert is the whole point of the warning.

2. **Level.** `WARN` is defensible and I'm not asking you to change it. But it's worth deciding deliberately: per the level table, WARN is "recoverable errors, degraded state", and a payments daemon silently trusting an unconfigured public RPC arguably clears the bar for ERROR ("critical failures requiring attention"). The counter-argument — that it's a legitimate dev-mode default — is fine; just make it a decision rather than a default.

daemon/src/configs/types.rs
```diff
@@ -596,3 +606,71 @@ pub struct SwapsConfig {
...
+    #[tracing_test::traced_test]
+    fn default_endpoints_are_announced_when_none_configured() {
+        let mut config = empty_chains_config();
+
+        config.set_default_chains_if_missing();
+
+        for chain in ChainType::iter() {
+            assert!(
+                config.chains[&chain]
+                    .get_random_requests_endpoint()
```
kirushik: **These tests are real — thank you for using `traced_test` rather than asserting on the substituted endpoints.** That distinction matters more than it looks: a test that only checks `chain_config.endpoints` would have passed with the `warn!` deleted, i.e. it would test the *old* behaviour. These genuinely observe the log buffer, and I verified the WARN lines are actually in it.

What they don't yet pin down — all cheap to add, and the reason I'd not call this done:

1. **Level isn't asserted.** The checks are global substring matches over the buffer, so a single ambiguous `INFO` event containing the same substrings would satisfy them. `tracing_test`'s `logs_assert` gives you the level; the repo already uses that pattern elsewhere.
2. **Count isn't asserted.** Nothing pins *exactly one event per affected chain*. Both a missing chain and a duplicated warning pass today.
3. **Field identity isn't asserted** — that the `chain` field actually names the chain that fell back, rather than a warning firing for the wrong one.
4. **Only the all-empty shape is covered.** Missing cases, which are also the enumeration of every way this can silently regress:
   - one chain configured, the other not (**and the reverse** — asymmetric bugs love this pair)
   - key present but the endpoint list is **empty** (`endpoints: []` vs key absent — these are different code paths through serde)
   - the loader-level path, not just the direct call, so env/JSON precedence is covered
   - the [`ZeroExApiConfig`](https://github.com/Kalapaja/Kalatori/blob/a82d0dbb8fa1101e7a91cfd192e8260d2c64acb3/daemon/src/configs/types.rs#L595) fallback from the other comment
5. **Nothing asserts a configured value is *absent* from the log.** Given endpoint URLs can carry credentials, a test that feeds a sentinel like `wss://user:hunter2@example.invalid` and asserts the sentinel never appears in the buffer would lock in the "only print hardcoded defaults" rule permanently. That's the one I'd most want.

A table-driven test over `(config shape) → (expected warnings)` would fold 1–4 into one place and make the coverage gaps visible as blank rows.

kirushik (commented):
## Review summary

Two independent passes, deliberately not shown each other's output: an **adversarial code pass** (GPT-5.6 `sol` via codex, high reasoning effort) and an **upstream-grounding pass** (Claude Opus with web access, which probed all four default endpoints live and executed the `config` crate against this repo's real source config).

**Verdict: the change is correct and the tests are real — but it's a half-fix, and the grounding pass found something bigger than the PR.**

### ✅ Verified good

- **The warning fires on the path that caused #333.** The chain default at [`configs/consts.rs:22`](https://github.com/Kalapaja/Kalatori/blob/a82d0dbb8fa1101e7a91cfd192e8260d2c64acb3/daemon/src/configs/consts.rs#L22) is `wss://polygon-bor-rpc.publicnode.com` — the exact endpoint from the incident. So this genuinely covers the motivating case.
- **The tests are not vacuous.** This was my main worry going in: a test that asserts only "the endpoints got substituted" tests the *old* behaviour, which already worked, and would pass with the `warn!` deleted. These don't — `tracing_test::traced_test` genuinely captures the log buffer and the WARN lines are really in it. Both passes confirmed independently, and both ran the tests.
- Emitted **once per affected chain**, not in a loop — no log-spam-then-filtered failure mode.
- CI clippy with the real flags (`--all-targets --all-features -Dwarnings`) exits 0; nightly `rustfmt` exits 0.
- **All four hardcoded default endpoints are alive.** Probed live on 2026-07-31 with a raw RFC-6455 client and HTTPS JSON-RPC: every one answered and streamed subscriptions. None is dead, deprecated or renamed, and both Polkadot defaults are current entries in polkadot-js `apps-config`. So the problem really is invisibility, not staleness.

### 🚨 The finding that outgrew the PR: chain endpoints can't be set by env var at all

The new warning tells operators to *"Configure your own endpoints for production use."* Via the mechanism [`README.md:85`](https://github.com/Kalapaja/Kalatori/blob/a82d0dbb8fa1101e7a91cfd192e8260d2c64acb3/README.md#L85) documents, **they can't.**

This wasn't reasoned about — it was executed, against `config 0.15.22` with this repo's exact source config:

- `KALATORI_CHAINS_CHAINS_POLYGON_ENDPOINTS` — parses fine, yields an **empty chains map**. Silently discarded.
- dotted forms — **hard error**, because `config` lowercases keys and `ChainType` has no `#[serde(rename_all)]`.

The JSON config file path still works. But the documented env-var override silently does nothing for chain endpoints, which is plausibly a *contributor to the three-month blindness in #333* rather than merely adjacent to it. Filed as #338; flagging here because it makes this PR's own advice unfollowable for anyone using env vars.

### ⚠️ Coverage is partial — see the inline comment

`ZeroExApiConfig::default()` hardcodes the **same free provider** with no warning, 13 lines above the new tests.

### 📌 One description claim didn't reproduce

The PR says the public endpoint "drops the websocket" after ~60 s. A 150-second run of unfiltered `eth_subscribe("logs")` against `wss://polygon-bor-rpc.publicnode.com` produced **101,975 notifications, no error frame, no close, no EOF.**

That doesn't refute three months of production data from a different vantage point — one probe from one location isn't the same evidence — but the ~60 s drop shouldn't be stated as an established property of the provider. Softening it to "observed in our deployment" would keep the argument intact without asserting something the next person can't reproduce.

Two related things worth knowing, both supporting the PR's underlying point:
- **Parity ships its public infrastructure against its own stated wishes**: *"should not be used by or linked to within any dapps, as we cannot guarantee response times or service level"* ([forum post](https://forum.polkadot.network/t/scaling-down-of-parity-s-public-infrastructure/4697) — note it names `rpc.polkadot.io`, not the Asset Hub host specifically).
- **dRPC is the only default publishing numbers**, and the fit is tighter than it looks: 20 CU/notification against a 120k CU/min cap that degrades to a 50,400 floor; measured USDC `Transfer` rate ~846/min → **~16.9k CU/min, ~34 % of the degraded floor**. Also a 2 s max timeout, batch capped at 3, and `eth_getLogs` capped at 10,000 entries — that last one directly constrains the gap-backfill remedy proposed in #333. ([dRPC rate limiting](https://drpc.org/docs/howitworks/ratelimiting))

So @kirushik's question in #333 — whether free public providers belong in compiled-in defaults at all — is well-founded, and this warning is the minimum viable answer rather than the whole one.

### 🧾 Pre-existing, found while reviewing

- **`allow_insecure_endpoints` is enforced on Asset Hub (subxt `from_url` vs `from_insecure_url`) but silently ignored on Polygon** (alloy `WsConnect`). A setting that means different things per chain is worse than one that means nothing.
- **No endpoint URL is ever parsed at config-load time**, so a typo surfaces as a connection failure at runtime rather than a config error at startup.
- [`docs/architecture.md:124`](https://github.com/Kalapaja/Kalatori/blob/a82d0dbb8fa1101e7a91cfd192e8260d2c64acb3/docs/architecture.md#L124) still describes chain endpoints as **mandatory**, which this PR leaves contradicted.

---

*Review method: two blind passes, worktree-isolated, no tracked file modified. Endpoint liveness and `config` crate behaviour were measured, not inferred.*

daemon/src/configs/types.rs
foxcool: Fixed — the 0x fallback is announced now, at the same point in startup as the chain warning.

`ZeroExApiConfig`'s URL moved into `DEFAULT_ZERO_EX_RPC_URL` in `consts.rs`, next to the `wss://` entry it shares a host with, and `swaps_config_with_prefix` calls `warn_if_zero_ex_rpc_is_public_default()` right after loading. Your framing was the deciding argument: partial coverage is worse than none here, because a clean log now reads as "I am fully configured".

I took the warning option rather than the other two you offered. Making the 0x RPC required, or making the client optional so it isn't constructed when unconfigured, are behaviour changes — and the `TODO` above the `Default` impl already asks for exactly that rework. Doing it inside a visibility fix would smuggle a startup-behaviour change into a PR nobody would think to look in.

One thing I chose deliberately: the warning fires whenever the URL *equals* the public default, not only when serde substituted it. By the time the config is loaded those two are indistinguishable, and the operator is on a free public node either way — which is the fact worth telling them.

Not covered, and I don't think it belongs here: `api_key` defaults to `""`, so an unconfigured 0x client is not merely degraded but broken at the first authenticated call. That's the `TODO`'s territory, not a visibility gap.

daemon/src/configs/types.rs
foxcool: All four addressed. The tests now go through `logs_assert` — the pattern already in `transfer_tracker.rs`, which you're right is the one that can actually observe a level — and match on level, message and the `chain` field **together**, in one predicate:

```rust
fn fallback_warnings_for(logs: &[&str], chain: ChainType) -> usize {
    let field = format!("chain={chain}");
    logs.iter()
        .filter(|log| log.contains(" WARN ") && log.contains(FALLBACK_MESSAGE) && log.contains(&field))
        .count()
}
```

That single change covers 1–3: an ambiguous INFO no longer satisfies anything, `assert_warned_once` / `assert_not_warned` pin the count at exactly one per affected chain, and a warning fired for the wrong chain fails on identity rather than passing on a substring.

On 4, the cases you enumerated, now seven tests in total:

- **one configured / other not, and the reverse** — `configuring_polygon_silences_polygon_only` and `configuring_asset_hub_silences_asset_hub_only`. You were right that the pair is the point: with a mutation that reports a fixed chain instead of the one that fell back, the Asset Hub direction still passes and only the Polygon direction fails.
- **`endpoints: []` vs key absent** — `an_explicitly_empty_endpoint_list_counts_as_unconfigured`, built by deserializing JSON rather than a struct literal, so it really does travel the explicit-empty-sequence path instead of `#[serde(default)]`.
- **loader-level** — `the_loader_announces_defaults_too` calls `chains_config_with_prefix` with a non-existent directory and a prefix unique to the test, so it exercises the file/env layer without being able to pick anything up from the environment.
- **`ZeroExApiConfig`** — two tests, defaulted and configured, per the other thread.

Verified by mutation, both directions:

- `warn!` → `info!`: 5 of 7 fail.
- `chain = %chain` → `chain = %ChainType::Polygon`: 4 fail, with both failure modes represented — `expected exactly one fallback WARN naming Polygon, found 2` (count) and `expected no fallback WARN naming Polygon, found 1` (identity).

kirushik (commented):
The swaps negative test still uses a bare `logs_contain`, the loader hook itself isn't exercised end-to-end anywhere in tests, and the PR body still asserts the "~60s websocket drop" as a provider property — despite the previous review already asking for softening.

foxcool:
All three done.

**The negative swaps test no longer opts out.** Both swaps tests now go through one level-aware helper, mirroring `fallback_warnings_for` on the chain side. You were right that this was the shape of the problem rather than one weak line: asymmetric rigour is how a pair rots, because the strict half keeps passing and nobody notices the other half stopped checking.

**The loader is covered end-to-end**, and this one was worth catching. `warn_if_zero_ex_rpc_is_public_default` was tested directly, but its only caller is `swaps_config_with_prefix` — and deleting the call there left the entire suite green. That is this PR's own bug, reintroduced through the one line nothing looked at. `the_swaps_loader_announces_the_public_default_too` now goes through the loader, matching what `the_loader_announces_defaults_too` already did for chains.

Verified by mutation, one per claim:

| mutation | result |
|---|---|
| `warn!` → `info!` | positive + loader fail; negative correctly does not |
| guard → `if true` | negative fails: `expected exactly 0 swaps fallback WARN, found 1` |
| call deleted from `configs.rs` | loader fails: `expected exactly 1 swaps fallback WARN, found 0` |

That third row is the one that previously produced nothing at all.

**The "~60s" claim is softened**, and you were right to ask twice — I carried the sentence over from the earlier body without re-reading it against the evidence. It now says the instance *was observed* reconnecting on roughly a 60-second period, one observation on one instance, not a property of the provider, and notes that nothing in this PR depends on the cause. The endpoint being invisible is the claim this PR actually rests on, and that one is directly checkable.

Rebased onto current `main` (post-#330/#334/#337). Clippy under `-Dwarnings`, fmt, 306 tests green.

PR #337, opened 2026-07-31
docs: describe the clippy lints that are actually enabled

foxcool:
Documentation only. `docs/conventions.md` and one line in `AGENTS.md`.

**Rewritten against current `main` rather than updated in place.** `main` has since grown the
panic gate and a much fuller `conventions.md` than this branch was written against, and most
of the original diff would have reverted it. What survived is the part still wrong there.

## The problem

`docs/conventions.md` said:

> **CI enforces** `RUSTFLAGS="-Dwarnings"` — all warnings are errors, including pedantic.

Three lines above a paragraph correctly stating that `pedantic` is enabled nowhere. `AGENTS.md`
carried the same claim without the contradiction to soften it.

This is not a harmless inaccuracy. It tells anyone writing code here that a class of mistakes
will be caught when nothing is watching for it. Concretely: `unused_self` is pedantic, and two
hash-building functions carried a `&self` they never used until it made them impossible to
unit test.

## The change

**Four explicit buckets**, so the section answers the question people actually arrive with —
*"will CI catch X?"*: explicit clippy lints, explicit rustc lint groups, compiler/clippy
defaults, and command-line escalation.

The second bucket was missing entirely. `[workspace.lints.rust]` sits in the same manifest as
the clippy table, and two of its four groups — `rust_2018_idioms`, `let_underscore` — contain
allow-by-default lints, so they add enforcement rather than restate the defaults. The
inheritance rule is stated too: a crate that drops `[lints] workspace = true` silently opts out
of everything, and nothing enforces that it does not.

**Where `-Dwarnings` applies**, precisely. It is set in exactly one place, the `cargo-clippy`
recipe, run by the PR clippy job. Build, test, coverage, integration and release jobs do not
set it; dependencies are capped at `warn` regardless. And it escalates lints that are *already
enabled* — it cannot turn on a lint that is off, which is why the list rather than the flag
determines what protects you. The old text got that backwards.

**"Not enabled" is no longer framed as a decision.** Git history says otherwise:

```
ed941bc3  2026-01-16  style: fix clippy and rustfmt warnings
  -arithmetic_side_effects = "deny"
  -pedantic = { level = "warn", priority = -1 }
abcaf132  2025-12-02  Add rustfmt config, format code
  -shadow_reuse / -shadow_same / -shadow_unrelated
```

The docs were accurate when written and quietly went stale — and the commit that made them
stale is titled "fix warnings" while actually silencing them by config. A regression with a
date reads differently to someone deciding whether to turn them back on than a deliberate
choice does.

**The `#[expect]` mechanism, stated correctly.** `#[expect]` is a scoped lint level
([RFC 2383](https://rust-lang.github.io/rfcs/2383-lint-reasons.html)): it raises the level for
its scope, so the lint runs there and the real violation fulfils the expectation. Clippy does
not evaluate disabled lints and discard the output — many never run at all. An earlier draft of
this PR had the observable behaviour right and the mechanism wrong.

## Measured, not guessed

Re-measured on current `main`, deduplicated by `(lint, file, line, column)` from
`--message-format=json` — heading-counting double-counts across the `bin` and `bin test`
targets, which is exactly the trap here.

**471 sites**, dominated by shadowing:

| lint | sites |
|---|---|
| `shadow_unrelated` | 149 |
| `shadow_reuse` | 52 |
| `unreadable_literal` | 45 |
| `arithmetic_side_effects` | 32 |
| `too_many_lines` | 25 |
| `doc_markdown` | 23 |
| `uninlined_format_args` | 19 |

So adopting them is a cleanup project, not a config change — which is why this PR fixes the
docs rather than the manifest.

The published command now carries Cargo's `--` separator (without it the command does not run
at all) and notes that `make setup` is a prerequisite, since `metadata.scale` and
`static/index.html` are generated inputs absent from a clean checkout — see #339.

Separately counted from source, because an `#[expect]` suppresses the warning that would
identify it: **28 of the 122 clippy expectations name a lint nothing enables.**

## Overlap to be aware of

#340 enables `arithmetic_side_effects` workspace-wide. If it lands first, the "what is not
enabled" section here needs the arithmetic rows removed; if this lands first, #340 needs to
update them. The two touch the same paragraphs of `conventions.md`.

## Verification

| Check | Result |
|---|---|
| clippy, `RUSTFLAGS="-Dwarnings"` | pass |
| tests | 292, unchanged — docs only |

Every factual claim re-verified on this tree rather than carried over: the two commits above,
the single `RUSTFLAGS` occurrence, which workflows invoke it, the 471 figure, and the
expectation counts.

### DIFF
--- a/docs/conventions.md
+++ b/docs/conventions.md
@@ -61,8 +61,10 @@ fails that job.
 
 The build, test, coverage, integration-test and release jobs do **not** set it,
 so a warning that only surfaces during a `cargo build` fails nothing.
-Dependencies are capped at `warn` regardless (`--cap-lints warn`), so nothing in
-the dependency graph can fail the build whatever the flag says.
+Dependencies are capped regardless, so nothing in the dependency graph can fail
+the build whatever the flag says. Nothing in this repo configures that — Cargo
+passes `--cap-lints warn` to every non-path dependency on its own, which is why
+`RUSTFLAGS` reaching the whole graph is harmless.
 
 Note also that `-Dwarnings` escalates lints that are *already enabled*. It
 cannot turn on a lint that is off — which is why the list above, not the flag,
--- a/docs/conventions.md
+++ b/docs/conventions.md
@@ -7,12 +7,15 @@
 - **rustfmt**: Nightly required (`cargo +nightly fmt --all`)
 - Self-named modules only (e.g., `chain.rs` + `chain/` directory) — **never `mod.rs`** (enforced by `mod_module_files` clippy lint). Rationale: better Git history, avoids file renaming issues.
 
-## Clippy Lints
+## Lints
 
-Workspace lints in root `Cargo.toml`:
+The question people arrive with is *"will CI catch X?"*, so the gate is
+described in the four parts that answer it.
+
+**1. Explicit clippy lints** — `[workspace.lints.clippy]` in the root
+`Cargo.toml`:
 
 ```toml
-[workspace.lints.clippy]
 allow_attributes = "deny"
 cargo_common_metadata = "warn"
 cast_possible_truncation = "warn"
@@ -20,12 +23,101 @@ ignored_unit_patterns = "warn"
 mod_module_files = "warn"
 ```
 
-**CI enforces** `RUSTFLAGS="-Dwarnings"` — all warnings are errors, including pedantic.
+Plus the eight panic-gate denials listed under [Panic Gate](#panic-gate) below,
+which live in the same table.
 
-There are **no** per-crate lint tables — `daemon`, `client` and `tools/cargo-bin`
-all carry only `[lints] workspace = true`, so the root block above plus the panic
-gate below is the complete set. In particular `pedantic` and
-`arithmetic_side_effects` are *not* enabled anywhere.
+**2. Explicit rustc lint groups** — `[workspace.lints.rust]`, in the same file
+and easy to miss:
+
+```toml
+future_incompatible = "warn"
+let_underscore = "warn"
+rust_2018_idioms = "warn"
+unused = "warn"
+```
+
+Two of these matter beyond the compiler's own defaults: `rust_2018_idioms` and
+`let_underscore` are groups containing allow-by-default lints, so they add
+enforcement rather than restating it.
+
+**3. Compiler and clippy defaults** — everything warn-by-default in `rustc` and
+`clippy`, always on.
+
+**4. Command-line escalation** — `RUSTFLAGS="-Dwarnings"`. This is narrower
+than it sounds, so it is spelled out in the next section.
+
+Both tables reach every crate through `[lints] workspace = true`, which
+`daemon`, `client` and `tools/cargo-bin` all carry. That inheritance is the
+load-bearing mechanism and it is invisible: a crate that omits the two lines
+silently opts out of *everything* above, and nothing enforces that it does not.
+There are **no** per-crate lint tables, so buckets 1–3 are the whole set.
+
+### Where `-Dwarnings` actually applies
+
+`RUSTFLAGS="-Dwarnings"` is set in exactly one place: the `cargo-clippy` recipe
+in the `Makefile`, invoked by the PR **clippy job**. It lints all targets and
+all features across the workspace, and anything an enabled lint reports there
+fails that job.
+
+The build, test, coverage, integration-test and release jobs do **not** set it,
+so a warning that only surfaces during a `cargo build` fails nothing.
+Dependencies are capped at `warn` regardless (`--cap-lints warn`), so nothing in
+the dependency graph can fail the build whatever the flag says.
+
+Note also that `-Dwarnings` escalates lints that are *already enabled*. It
+cannot turn on a lint that is off — which is why the list above, not the flag,
+determines what protects you.
+
+### What is not enabled, and why the docs used to say otherwise
+
+`pedantic`, `arithmetic_side_effects` and the `shadow_*` family are **off**.
+
+They were not declined after deliberation — they were configured and then
+removed, and the documentation simply went stale:
+
+- `pedantic` and `arithmetic_side_effects` were removed on 2026-01-16 in
+  `ed941bc`, titled *"style: fix clippy and rustfmt warnings"* — a commit that
+  silenced the warnings by config rather than fixing them.
+- The `shadow_*` family went the same way in `abcaf13` (2025-12-02).
+
+Turning them back on today produces **471 warnings** across the workspace,
+deduplicated by site — so adopting them is a cleanup project, not a config
+change. The bulk is shadowing (149 `shadow_unrelated`, 52 `shadow_reuse`), then
+45 unseparated long literals, 32 unchecked arithmetic operations, 25
+over-long functions, 23 missing doc backticks and 19 inlinable `format!` args.
+
+Re-measure with:
+
+```bash
+cargo clippy --all-targets --all-features -- \
+  -W clippy::pedantic \
+  -W clippy::arithmetic_side_effects \
+  -W clippy::shadow_reuse -W clippy::shadow_same -W clippy::shadow_unrelated
+```
+
+The `--` separator is required: lint flags belong to `clippy-driver`, and Cargo
+rejects them without it. Run `make setup` first — `metadata.scale` and
+`static/index.html` are generated inputs, absent from a clean checkout, and the
+daemon aborts before finishing the lint pass without them ([#339](https://github.com/Kalapaja/Kalatori/issues/339)).
+
+### `#[expect]` on a lint nothing enables
+
+`#[expect(clippy::too_many_lines)]` and friends sit in the tree without tripping
+`unfulfilled_lint_expectations`, even though `pedantic` is off. Of the 122
+clippy expectations in first-party code, **28 name a lint nothing enables**,
+across six of them: `arithmetic_side_effects` (13), `too_many_lines` and
+`cast_sign_loss` (4 each), `struct_field_names` and `module_name_repetitions`
+(3 each), and `unused_self`.
+
+That works because `#[expect]` is a *scoped lint level* ([RFC 2383](https://rust-lang.github.io/rfcs/2383-lint-reasons.html)):
+it raises the lint's level for its scope, so the lint runs there even though
+nothing enables it globally, and the genuine violation fulfils the expectation.
+Clippy does not evaluate disabled lints and discard the output — many never run
+at all. Adding an expectation where the code does *not* violate the lint still
+warns.
+
+So these annotations mark real violations of lints nothing enforces. Useful as
+documentation; not evidence that anything is checking.
 
 ## Panic Gate
 
@@ -129,17 +221,17 @@ In order of preference:
   typed error over extending the marker to new code.
 
 - **Arithmetic is not gated.** `clippy::arithmetic_side_effects` would catch
-  overflow, division by zero and `Decimal` panics; roughly two dozen daemon
-  sites trip it today. Division by zero panics unconditionally; overflow
-  currently *wraps silently* in release, which for a daemon computing payment
-  amounts is arguably worse than panicking.
+  overflow, division by zero and `Decimal` panics; 32 sites trip it today (see
+  the measurement above), plus 13 already carrying an `#[expect]`. Division by
+  zero panics unconditionally; overflow currently *wraps silently* in release,
+  which for a daemon computing payment amounts is arguably worse than panicking.
 
 - **`[profile.release]` is in `daemon/Cargo.toml`, a workspace member, so Cargo
   ignores it** — it prints `profiles for the non root package will be ignored`
   on every build. `panic = "abort"`, `overflow-checks`, `lto`, `strip` and
   `codegen-units` are therefore *not* in effect. Moving the block to the
   workspace root should be sequenced **after** the arithmetic gate above:
-  enabling `overflow-checks` while two dozen unguarded sites remain would turn
+  enabling `overflow-checks` while those sites remain unguarded would turn
   silent wrapping into live panics, i.e. create the very DoS class this gate
   exists to close.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,7 +6,7 @@ Self-hosted, non-custodial blockchain payment gateway daemon for Polkadot Asset 
 
 - **Metadata regen required** when updating subxt or connecting to new chain version: `make install-subxt-cli && make download-node-metadata-ci`
 - **Version sync**: subxt-cli must match subxt in Cargo.toml (both 0.44), sqlx-cli must match sqlx (both 0.8) — versions pinned in `Makefile`
-- **Clippy strict**: CI runs `RUSTFLAGS="-Dwarnings"` — all warnings are errors, including pedantic lints
+- **Clippy**: CI runs `RUSTFLAGS="-Dwarnings"`, so every *enabled* lint is a build failure — but the enabled set is small and **`pedantic` is not in it**. See [docs/conventions.md](docs/conventions.md) before assuming a lint protects you
 - **Never use `mod.rs`** — self-named modules only (enforced by `mod_module_files` clippy lint)
 - **Nightly rustfmt**: `cargo +nightly fmt --all`
 - **SQLite >= 3.47.0** required at runtime (see README.md for build-from-source instructions)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,7 +6,7 @@ Self-hosted, non-custodial blockchain payment gateway daemon for Polkadot Asset 
 
 - **Metadata regen required** when updating subxt or connecting to new chain version: `make install-subxt-cli && make download-node-metadata-ci`
 - **Version sync**: subxt-cli must match subxt in Cargo.toml (both 0.44), sqlx-cli must match sqlx (both 0.8) — versions pinned in `Makefile`
-- **Clippy**: CI runs `RUSTFLAGS="-Dwarnings"`, so every *enabled* lint is a build failure — but the enabled set is small and **`pedantic` is not in it**. See [docs/conventions.md](docs/conventions.md) before assuming a lint protects you
+- **Clippy**: the PR clippy job — and only that job — runs `RUSTFLAGS="-Dwarnings"`, so every *enabled* lint fails it. The enabled set is small and **`pedantic` is not in it**. See [docs/conventions.md](docs/conventions.md) before assuming a lint protects you
 - **Never use `mod.rs`** — self-named modules only (enforced by `mod_module_files` clippy lint)
 - **Nightly rustfmt**: `cargo +nightly fmt --all`
 - **SQLite >= 3.47.0** required at runtime (see README.md for build-from-source instructions)


docs/conventions.md
```diff
@@ -19,12 +19,31 @@ ignored_unit_patterns = "warn"
...
+Every member inherits these via `[lints] workspace = true`. There are **no per-crate
+lints** — that list plus rustc's defaults is the whole gate.
+
+**CI enforces** `RUSTFLAGS="-Dwarnings"`, so anything the enabled lints report fails the
+build.
+
+### What is deliberately not enabled
+
+`pedantic`, `arithmetic_side_effects` and the `shadow_*` family are **off**. This document
+used to claim they were enabled per-crate; they never were. Turning them on today produces
```
kirushik: **"they never were" is refuted by this repo's own git history — and the difference matters, because this is a regression, not a documentation error.**

Both review passes found the same thing independently:

- `pedantic = { level = "warn", priority = -1 }` and `arithmetic_side_effects = "deny"` **were** configured in the root `Cargo.toml`, and were removed on **2026-01-16 in `ed941bc`**, titled *"style: fix clippy and rustfmt warnings"*.
- The `shadow_*` family was present from `major: init` and dropped in `abcaf13`.
- The deleted lines match the old documentation **character-for-character** — which is almost certainly why the docs said what they said: they were accurate when written, and nobody updated them when the config changed.

So the old docs weren't a fantasy someone invented; they were a true statement that quietly went stale. And the commit that made them stale is titled "fix warnings" while actually *silencing* them by config.

That reframes the whole section. As written, the heading **"What is deliberately not enabled"** asserts an intent the history contradicts — it tells a future reader that someone weighed these lints and declined them, when what actually happened is that a cleanup commit turned them off in passing and the docs drifted for six months.

**Suggested rewrite** of the sentence:

> ~~This document used to claim they were enabled per-crate; they never were.~~
>
> These were enabled in the root `Cargo.toml` until `ed941bc` (2026-01-16, *"style: fix clippy and rustfmt warnings"*), which removed them rather than fixing the violations; the `shadow_*` family went the same way in `abcaf13`. This document was not updated at the time, which is why it described them as enabled until now.

And retitle the section to something factual — "What is not currently enabled" — unless you want to make it deliberate, in which case that's a decision to take explicitly rather than to document retroactively. See the review body for why `arithmetic_side_effects` in particular is worth taking that decision on: clippy's own docs use this repo's money type as the motivating example, and the runtime `overflow-checks` backstop turns out not to be applied either.

A one-line "how we got here" is also just *useful*: the next person who wonders why a money daemon has no arithmetic lint gets an answer instead of a mystery.

docs/conventions.md
```diff
@@ -19,12 +19,31 @@ ignored_unit_patterns = "warn"
...
 ```
 
-**CI enforces** `RUSTFLAGS="-Dwarnings"` — all warnings are errors, including pedantic.
-
-Per-crate lints (in `daemon/Cargo.toml`):
-- `pedantic = { level = "warn", priority = -1 }`
-- `arithmetic_side_effects = "warn"`
-- `shadow_reuse`, `shadow_same`, `shadow_unrelated` = "warn"
+Every member inherits these via `[lints] workspace = true`. There are **no per-crate
+lints** — that list plus rustc's defaults is the whole gate.
```
kirushik: **"that list plus rustc's defaults is the whole gate" isn't quite the whole gate — there's a second lint table in the same file.**

[`Cargo.toml:22-26`](https://github.com/Kalapaja/Kalatori/blob/2e74320d9f3915bd7a8cf612cbd00f7ed9f097ef/Cargo.toml#L22-L26) has a `[workspace.lints.rust]` table alongside the clippy one:

```toml
future_incompatible
let_underscore
rust_2018_idioms
unused
```

These are inherited by every member through the same `[lints] workspace = true`, and **two of them (`rust_2018_idioms`, `let_underscore`) are groups containing allow-by-default lints** — i.e. they're not subsumed by "rustc's defaults", they genuinely add enforcement (`rustc -W help` confirms). A PR whose title is "the lints that are actually enabled" is the one place this shouldn't be missing.

**Suggested:** split the section into four explicit buckets, which also makes the document answer the question people actually arrive with — *"will CI catch X?"*:

1. **Explicit clippy lints** — the five in `[workspace.lints.clippy]`
2. **Explicit rustc lint groups** — the four in `[workspace.lints.rust]`
3. **Compiler and clippy defaults** — warn-by-default, always on
4. **Command-line escalation** — what `-Dwarnings` does to the above, and where it applies (see the other comment; it's narrower than the current text implies)

Worth stating the inheritance rule explicitly too, since it's the load-bearing mechanism and it's invisible: a crate that omits `[lints] workspace = true` silently opts out of *everything* above. All four manifests currently have it — I checked — but nothing enforces that for a future crate, so it's exactly the kind of thing a conventions doc should name.

docs/conventions.md
```diff
@@ -19,12 +19,31 @@ ignored_unit_patterns = "warn"
...
+
+**CI enforces** `RUSTFLAGS="-Dwarnings"`, so anything the enabled lints report fails the
+build.
```
kirushik: **This overstates the scope. `-Dwarnings` is set in exactly one place, and it doesn't cover most of what "CI" implies.**

Verified by reading the workflows and confirming with `cargo clippy -vv`:

- `RUSTFLAGS="-Dwarnings"` is set **only** by the `cargo-clippy` make recipe, invoked from the PR clippy job.
- The **build, test, merge-to-main and release jobs do not set it.** A warning that only appears during a `cargo build` never fails anything.
- **Dependencies are capped at `warn` regardless** (`--cap-lints warn`), so nothing in the dependency graph can fail the build no matter what the flag says.

The current sentence — *"CI enforces `RUSTFLAGS="-Dwarnings"`, so anything the enabled lints report fails the build"* — reads as a blanket property of CI. In practice it's one job, one command, workspace crates only.

**Suggested:**

> The PR **clippy job** runs `make cargo-clippy`, which sets `RUSTFLAGS="-Dwarnings"` and lints all targets and all features across the workspace. Anything an enabled lint reports there fails that job. Build, test and release jobs do **not** set it, and dependency lints are capped at `warn` — so a warning outside the clippy job's scope will not fail anything.

Worth being precise here specifically because the *old* text's version of this claim was wrong in a more serious way (it said `-Dwarnings` made pedantic an error — doubly wrong, since `-Dwarnings` escalates warnings that are *already enabled* and cannot enable a lint that is off). This PR is right to fix that; it's just worth landing the precise version rather than a differently-imprecise one.

While you're in here: [`AGENTS.md:9`](https://github.com/Kalapaja/Kalatori/blob/2e74320d9f3915bd7a8cf612cbd00f7ed9f097ef/AGENTS.md#L9) now carries the same claim in compressed form (*"CI runs `RUSTFLAGS="-Dwarnings"`, so every enabled lint is a build failure"*). That's much better than before — the "but the enabled set is small and `pedantic` is not in it" clause does real work — but "build failure" should probably be "fails the clippy job" for the same reason.

docs/conventions.md
```diff
@@ -19,12 +19,31 @@ ignored_unit_patterns = "warn"
...
+
+### What is deliberately not enabled
+
+`pedantic`, `arithmetic_side_effects` and the `shadow_*` family are **off**. This document
+used to claim they were enabled per-crate; they never were. Turning them on today produces
+**461 warnings** (measured with `cargo clippy --all-targets --all-features -W
+clippy::pedantic -W clippy::arithmetic_side_effects -W clippy::shadow_reuse -W
+clippy::shadow_same -W clippy::shadow_unrelated`), so adopting them is a cleanup project,
```
kirushik: **This command doesn't run — it's missing Cargo's `--` separator, so anyone who copies it to check your numbers gets an error instead.**

```console
$ cargo clippy --all-targets --all-features -W clippy::pedantic ...
error: unexpected argument '-W' found
```

Lint flags have to go after `--`, where they're passed through to `clippy-driver` rather than parsed by Cargo:

```bash
cargo clippy --all-targets --all-features -- \
  -W clippy::pedantic \
  -W clippy::arithmetic_side_effects \
  -W clippy::shadow_reuse -W clippy::shadow_same -W clippy::shadow_unrelated
```

Small thing, but this command is the *evidence* for the 461 figure — it's the one line in the section a skeptical reader will actually execute, so it's the worst one to have broken.

**On the number itself:** neither review pass could confirm it, and not because of the separator. `metadata.scale` and `static/index.html` are gitignored generated inputs absent from a clean checkout, so **every daemon clippy invocation exits 101 before finishing** — with or without the flags. Partial output before the abort showed 74 warning headings (dominated by the 44 unreadable literals you mention, which is consistent with your breakdown); per-package runs on `kalatori-client` and `cargo-bin` completed cleanly and did produce pedantic warnings.

So the figure is plausible and probably right — it just isn't reproducible by a reader from a fresh clone, which undercuts the point of publishing it. Two options:

1. Note the prerequisite inline (*"after `make setup`, which generates `metadata.scale` and the frontend"*), or
2. Quote a per-crate breakdown for the crates that lint standalone, which anyone can reproduce immediately.

The underlying "repo can't be linted or tested from a clean clone" problem is filed as #339 — it's not this PR's to fix, but it *is* why this number can't be checked.

docs/conventions.md
```diff
@@ -19,12 +19,31 @@ ignored_unit_patterns = "warn"
...
+
+Two consequences worth knowing:
+
+- **`#[expect(clippy::…)]` on a disabled lint still works.** Clippy evaluates the lint even
+  at `allow` level and only suppresses the output, so the expectation is fulfilled by the
+  real violation. That is why `#[expect(clippy::too_many_lines)]` and friends appear in the
+  tree without tripping `unfulfilled_lint_expectations` — they mark genuine violations of
+  lints nothing enforces. Adding one where the code does *not* violate the lint does warn.
+- **Nothing catches what those lints catch.** `unused_self`, for one, is pedantic: two
+  hash-building functions carried a `&self` they never used until it made them untestable.
```
kirushik: **The conclusion is right and genuinely useful — the stated mechanism isn't. Worth fixing because this paragraph is teaching a mental model.**

The claim *"Clippy evaluates the lint even at `allow` level and only suppresses the output"* is not how it works. `expect` is a **scoped lint level** ([RFC 2383](https://rust-lang.github.io/rfcs/2383-lint-reasons.html)), and clippy demonstrably short-circuits on level — `module_style.rs`, for instance, returns early when the level is `Allow`. Lints aren't evaluated-then-suppressed; many are never run at all.

The *observable behaviour* you describe is nonetheless correct, and both passes confirmed it empirically — `#[expect(clippy::too_many_lines)]` and friends do sit in the tree without tripping `unfulfilled_lint_expectations`, across 44 `#[expect]`s over 11 lints. The mechanism is that `#[expect]` **raises the lint's level in that scope**, so the lint runs *there* even though it's off globally, and the expectation is fulfilled by the real violation.

**Suggested rewrite:**

> **`#[expect(clippy::…)]` on a globally-disabled lint still works.** The attribute sets the lint level for its scope, so the lint runs there even though nothing enables it globally, and the expectation is fulfilled by the genuine violation. That's why `#[expect(clippy::too_many_lines)]` and friends appear in the tree without tripping `unfulfilled_lint_expectations` — they mark real violations of lints nothing enforces. Adding one where the code does *not* violate the lint does warn.

Same paragraph, accurate mechanism, and it now explains *why* rather than asserting a behaviour.

---

On **"Nothing catches what those lints catch"** just below: too absolute, and it undersells the current gate.

- `cast_possible_truncation` **is** enabled, and it covers the `as`-cast half of what people assume `arithmetic_side_effects` does (that lint does *not* fire on casts).
- Compiler and default-clippy checks overlap with subsets of pedantic.
- The 44 local `#[expect]`s evaluate several otherwise-disabled lints at their sites, per the paragraph above.

Suggest narrowing to something like *"violations unique to these groups are not gated anywhere in CI"* — true, and it keeps the reader from concluding there's zero protection.

The `unused_self` example is a good one, by the way — "two hash-building functions carried a `&self` they never used until it made them untestable" is concrete and memorable, and it's exactly the kind of thing that makes a conventions doc worth reading. It's also nicely timed: #334 is the PR that removes those `&self` receivers so the hashes can be pinned by known-answer tests.

kirushik (commented):
## Review summary

Two independent passes, deliberately not shown each other's output: an **adversarial pass** (GPT-5.6 `sol` via codex, high reasoning effort, verifying every assertion against the build configuration and running the commands) and an **upstream-grounding pass** (Claude Opus with web access, reading clippy's lint declarations from `rust-lang/rust` at tag **1.95.0** to match `rust-toolchain.toml`, plus the Cargo Book and RFC 2383).

**The central claim is correct and this correction is overdue.** `pedantic`, `arithmetic_side_effects` and the `shadow_*` family are genuinely not enabled; the old text was wrong; documenting reality is the right instinct. Confirmed independently: all four crate manifests carry `[lints] workspace = true` (none opts out), there is no `clippy.toml`, and no crate-level `#![warn]`/`#![deny]`/`#![allow]` attributes exist. `mod_module_files` really is configured, really does become fatal in the PR clippy job, and no `mod.rs` exists.

**But both passes independently concluded the replacement text is also wrong in several places** — and a document whose entire purpose is "be the accurate record" has a higher bar than the one it replaces. Details inline; the big one is that "they never were" is refuted by this repo's own history.

### 🔗 Coordinate with #330 on MSRV

The MSRV contradiction — `AGENTS.md` and `docs/conventions.md` say **1.88**, `daemon/Cargo.toml` declares **1.91**, `rust-toolchain.toml` pins builds to **1.95.0** — sits within a few lines of what this PR edits, in the same files, and is left standing.

**#330 already fixes it**, moving `rust-version` to `[workspace.package]` and updating `AGENTS.md`, `CONTRIBUTING.md` and `docs/conventions.md` to 1.91 (verified correct: scanning all 611 lockfile packages gives a true floor of exactly 1.91, set by alloy 2.1.1). Whichever of the two lands second will conflict here. Simplest resolution: let #330 own the MSRV numbers, and this PR own the lint text.

### ✅ One suspicion I had, refuted — the nightly-rustfmt requirement is NOT stale

I expected `docs/conventions.md:6`'s nightly requirement to be another stale claim. It isn't: running stable rustfmt (1.97.1) against this repo's `rustfmt.toml` **rejects 7 of the 12 options** as nightly-only. The line is correct as written; no action needed. Recording it so nobody else re-opens it.

### 🚨 The substantive issue behind the documentation

This PR asks "what do the docs say?" The more important question is "what should they say?", and the evidence points one way:

1. **`arithmetic_side_effects` matters more for *this* codebase than the generic case.** Clippy's own documentation uses **`rust_decimal::Decimal` as its motivating third-party example** ([`operators/mod.rs` @ 1.95.0](https://github.com/rust-lang/rust/blob/1.95.0/src/tools/clippy/clippy_lints/src/operators/mod.rs)) — this repo's money type. Confirmed empirically that it fires on `Decimal + Decimal`. It does *not* cover `as` casts (that's `cast_possible_truncation`, which **is** enabled), floats, `checked_*`, or const contexts — so the loss is narrower than "all arithmetic", but it lands exactly on the money path.
2. **Twelve `#[expect(clippy::arithmetic_side_effects)]` attributes sit in the tree as inert markers**, implying an enforcement that doesn't exist.
3. **The runtime backstop is also missing.** [`daemon/Cargo.toml:141`](https://github.com/Kalapaja/Kalatori/blob/2e74320d9f3915bd7a8cf612cbd00f7ed9f097ef/daemon/Cargo.toml#L141) declares `overflow-checks = true`, but **Cargo silently ignores `[profile.*]` in a non-root workspace member** ([Cargo Book](https://doc.rust-lang.org/cargo/reference/profiles.html)). Verified from the actual `rustc` invocation via `cargo build --release -vv`: no overflow-check flag. Independently confirmed by reviewers on #330 and #334 as well.

Together: **a payment daemon computing invoice totals, refunds, token-unit conversions and gas costs has neither the compile-time lint nor the runtime overflow check its own configuration claims.** Concrete arithmetic risk was confirmed in invoice totals/refunds, token-unit conversion and Polygon gas-cost calculation.

That's a decision for you rather than something this PR should assume — but "restore the lints" and "document their absence" are different answers, and right now the PR quietly picks the second. A fix for the profile half is being prepared separately.

### 📌 Verification caveat, stated honestly

**Neither pass could verify the "461 warnings" figure**, and neither could run clippy on the daemon at all: `metadata.scale` and `static/index.html` are gitignored generated inputs absent from a clean checkout, so every daemon clippy invocation exits 101. Partial output before the abort showed 74 warning headings dominated by 44 unreadable literals; per-package runs on `kalatori-client` and `cargo-bin` completed cleanly. The number may well be right — it just can't be reproduced from a fresh clone, which is itself worth fixing — filed as #339.

CI on this branch is **not** fully green, incidentally: `cargo-deny (advisories)` fails on [RUSTSEC-2026-0220](https://rustsec.org/advisories/RUSTSEC-2026-0220) (`ruint 1.18.0`) inherited from `main` — #330 is the fix. Unrelated to this PR, but don't read the branch as clean.

---

*Review method: two blind passes, worktree-isolated, no tracked file modified. Behavioural claims about clippy and Cargo were tested in scratch crates rather than reasoned about.*

foxcool:
Cross-link from #330, so neither of us re-fixes the other's line.

#330 changes the MSRV statements in the same two files — `AGENTS.md` and `docs/conventions.md` — from 1.88 to **1.91**, the measured floor of the lockfile, and moves the declaration to `[workspace.package]` so all three members inherit it.

This PR does not touch those lines, so there is no contradiction to resolve twice. But the hunks are near each other, so whichever lands second may want a trivial rebase. #330 is currently ahead — it is queued behind #334 by merge order.

Also worth knowing while this one is about lint-config drift: `clippy::arithmetic_side_effects` is being restored in #340, and the `[profile.*]` tables that Cargo silently ignores are moved there too. Whatever this PR ends up documenting should describe the state after that lands, or explicitly say it does not.

docs/conventions.md
foxcool: Verified independently and adopted — you're right on both the fact and why it matters.

```
ed941bc3  2026-01-16  style: fix clippy and rustfmt warnings
  -arithmetic_side_effects = "deny"
  -pedantic = { level = "warn", priority = -1 }
abcaf132  2025-12-02  Add rustfmt config, format code
  -shadow_reuse = "warn"  -shadow_same = "warn"  -shadow_unrelated = "warn"
```

The section no longer presents "off" as a decision. It now says these were configured until those two commits, notes that `ed941bc` is titled "fix warnings" while actually silencing them by config, and states plainly that the docs were true when written and nobody updated them.

Your framing is what changed my mind about the heading: **"deliberately not enabled" asserts an intent the history contradicts.** A reader deciding whether to turn them back on is answering a different question depending on which story they get — "someone weighed this and declined" versus "this regressed six months ago and nobody noticed". The second is the true one and it argues the opposite way.

Note this branch was rewritten against current `main` rather than updated in place: `main` has since grown the panic gate and a far fuller `conventions.md` than this PR was written against, and most of the original diff would have reverted it. What survived is the part still wrong there — `-Dwarnings` "all warnings are errors, including pedantic", sitting three lines above a paragraph correctly saying pedantic is enabled nowhere.

docs/conventions.md
foxcool: Fixed, and restructured into the four buckets you proposed — they do answer the question people actually arrive with.

`[workspace.lints.rust]` was the outright omission, and your point about it stands: `rust_2018_idioms` and `let_underscore` are groups containing allow-by-default lints, so they genuinely add enforcement and are not subsumed by "rustc's defaults". Both are now their own bucket with that noted.

The inheritance rule is spelled out too, with your reasoning for why it belongs in the doc at all — it is load-bearing and invisible, and a crate that drops `[lints] workspace = true` silently opts out of everything. I checked the four manifests again and they all still carry it; the doc says so, and says nothing enforces it.

One structural change beyond what you asked: the eight panic-gate denials live in the same `[workspace.lints.clippy]` table, so bucket 1 points at them explicitly rather than letting the code block imply the table has five entries.

docs/conventions.md
foxcool: Fixed, using your wording almost verbatim — the precise version is worth landing.

Confirmed the mechanism myself before writing it:

```
$ grep -rn "Dwarnings\|RUSTFLAGS" Makefile .github/ .cargo/
Makefile:97:	RUSTFLAGS="-Dwarnings" cargo clippy --all-targets --all-features
```

One place. `_job-clippy.yml` is the only workflow that runs `make cargo-clippy`; the test, coverage, integration-test and release jobs run `make cargo-test` / `make generate-coverage-report` / `make run-test-examples` and set nothing.

The doc now says clippy job, all targets, all features, workspace crates only, and that dependencies are capped at `warn` regardless.

Your parenthetical is in there as its own paragraph, because it is the part that actually prevents the mistake: `-Dwarnings` escalates lints that are *already enabled* and cannot enable one that is off — so the list, not the flag, determines what protects you. The old text got this doubly wrong and it is worth stating positively rather than just deleting.

docs/conventions.md
foxcool: Both fixed, and I re-measured rather than carrying the number forward.

The `--` separator is in. You were right that this was the worst line to have broken — it is the one a sceptical reader executes, and it is the evidence for the figure.

**The number is now 471, not 461.** Measured on current `main` (0.9.4), deduplicated by `(lint, file, line, column)` from `--message-format=json` rather than counting warning headings, which double-counts across the `bin` and `bin test` targets — the raw per-target totals are 206 + 436 with 196 duplicates flagged, so heading-counting is exactly the trap here. Breakdown:

| lint | sites |
|---|---|
| `shadow_unrelated` | 149 |
| `shadow_reuse` | 52 |
| `unreadable_literal` | 45 |
| `arithmetic_side_effects` | 32 |
| `too_many_lines` | 25 |
| `doc_markdown` | 23 |
| `uninlined_format_args` | 19 |

Consistent with the shape you saw in the partial output, and the conclusion is unchanged: cleanup project, not a config change.

On reproducibility — your diagnosis was right and the cause is `metadata.scale` and `static/index.html` being gitignored generated inputs. The command runs to completion here because `make setup` has been run; from a clean clone it still aborts at 101. I took your option 1: the prerequisite is noted inline, with a pointer to #339, which is the issue that should actually fix it.

The arithmetic figure is now cross-referenced from the Backlog section too, which previously said "roughly two dozen" — 32 sites trip the lint, plus 13 more already carrying an `#[expect]`.

docs/conventions.md
foxcool: Corrected — you're right, and thank you for the RFC pointer, which is what made the distinction land.

The paragraph now says `#[expect]` is a scoped lint level that **raises the lint's level for its scope**, so the lint runs there even though nothing enables it globally, and the genuine violation fulfils the expectation. The "evaluates at allow level and suppresses the output" claim is gone, with an explicit line saying clippy does not do that — many lints never run at all, `module_style.rs` returning early on `Allow` being the concrete case.

Worth recording that the observable behaviour I originally verified was real; the mechanism I inferred from it was not. Two different things, and only one of them was tested.

I re-counted on this tree rather than reusing 44/11: **28 of the 122 clippy expectations in first-party code name a lint nothing enables**, across six — `arithmetic_side_effects` (13), `too_many_lines` and `cast_sign_loss` (4 each), `struct_field_names` and `module_name_repetitions` (3 each), `unused_self` (1). The rest name lints that *are* enabled, mostly the panic gate, and those are a different thing entirely — they satisfy a live gate rather than annotating an absent one.

One methodological note, since it bit me: these cannot be counted from the warning output, because an `#[expect]` suppresses the very warning that would identify it. They have to be counted from the source and classified by group.

docs/conventions.md
```diff
@@ -7,25 +7,117 @@
...
+
+`RUSTFLAGS="-Dwarnings"` is set in exactly one place: the `cargo-clippy` recipe
+in the `Makefile`, invoked by the PR **clippy job**. It lints all targets and
+all features across the workspace, and anything an enabled lint reports there
+fails that job.
+
+The build, test, coverage, integration-test and release jobs do **not** set it,
+so a warning that only surfaces during a `cargo build` fails nothing.
+Dependencies are capped at `warn` regardless (`--cap-lints warn`), so nothing in
+the dependency graph can fail the build whatever the flag says.
```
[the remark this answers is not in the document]
foxcool: Reworded — you're right that the sentence implied configuration, and wrong about the underlying behaviour, so both halves are worth stating.

`make cargo-clippy` does not pass `--cap-lints`, and neither does anything else in the repo. But the capping is real: **Cargo applies it itself** to every non-path dependency. Verified rather than argued — a throwaway crate with a single registry dependency, `RUSTFLAGS="-Dwarnings"`, no `--cap-lints` anywhere in the manifest:

```
$ cargo build -vv | grep -o "cap-lints [a-z]*"
cap-lints warn
```

So the fix is attribution, not deletion. The line now says nothing here configures it, that Cargo does it on its own, and — the part that makes it worth a sentence at all — that this is *why* `RUSTFLAGS` reaching the whole dependency graph is harmless.

kirushik (approved):
:shipit:

PR #392, opened 2026-08-07
fix(swaps): verify settlement before a swap can direct a refund

kirushik:
Swap endpoints are unauthenticated, and a refund with no explicit destination preferred the earliest completed incoming swap's `from_address`. Nothing tied that swap to the invoice being refunded: an attacker who knew an invoice id could create a swap with their own `from_address`, submit a publicly-obtainable transaction hash, and collect someone else's refund. For normal 0x the status check was `get_transaction_receipt(hash).status()`, so any successful Polygon transaction did it.

Only two executors can reach an incoming swap — `detect` assigns 0x normal same-chain and Across cross-chain — so those are what this covers. Gasless is outgoing-only and already excluded by the refund query's direction filter; Bungee is unreachable, as nothing constructs it.

0x normal now has its Polygon receipt decoded before completion: the settlement must carry an ERC-20 transfer of the stored token, on the stored chain, to the stored recipient, or the swap is failed with the detail an operator needs. Across keeps completing exactly as before — its stored reference identifies a source-chain deposit, not the Polygon fill, so nothing here can verify it — but it never earns the marker, and the refund-eligibility query requires one. Unverifiable settlements therefore fall through to manual handling rather than paying out on evidence nobody checked. The detector separately refuses any swap whose destination is not the invoice's payment address.

The amount is deliberately advisory: the stored swap carries neither the quoted slippage allowance nor the destination token's decimals, so no honest lower bound exists yet. A legitimate fill landing slightly short completes and warns rather than stranding the payment; recipient, token and chain are what close the theft.

The marker lives in `error_message` to avoid a migration. That is sound only because `Failed` is terminal in the schema trigger and the two writers of that column are this completion path and the failure path, which always sets `Failed` — so a `Completed` row cannot carry a forged marker. It is not exposed in `PublicSwap`. A dedicated column belongs in the next migration.

### DIFF
--- a/daemon/src/chain_client.rs
+++ b/daemon/src/chain_client.rs
@@ -21,6 +21,7 @@ use uuid::Uuid;
 use crate::types::{
     ChainType,
     GeneralTransactionId,
+    SwapChainType,
     TransferInfo,
 };
 
@@ -102,6 +103,39 @@ pub struct GeneralChainTransfer {
     pub transaction_hash: Option<String>,
     // TODO: use DateTime<Utc> for consistency
     pub timestamp: u64, // milliseconds since epoch
+}
+
+/// An ERC-20 transfer decoded from a destination-chain transaction receipt.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SettlementTransfer {
+    pub token_address: String,
+    pub recipient_address: String,
+    pub amount_units: Option<u128>,
+}
+
+/// Destination-chain evidence used to independently confirm swap settlement.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SwapSettlement {
+    pub chain: SwapChainType,
+    pub transfers: Vec<SettlementTransfer>,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+pub enum SettlementVerificationError {
+    #[error("invalid transaction hash")]
+    InvalidTransactionHash,
+    #[error("destination-chain receipt is not available")]
+    ReceiptUnavailable,
+}
+
+/// Reads destination-chain settlement evidence independently of swap providers.
+#[cfg_attr(test, mockall::automock)]
+#[trait_variant::make(Send)]
+pub trait SwapSettlementVerifier: Sync {
+    async fn get_swap_settlement(
+        &self,
+        transaction_hash: &str,
+    ) -> Result<SwapSettlement, SettlementVerificationError>;
 }
 
 impl GeneralChainTransfer {
--- a/daemon/src/chain_client/polygon.rs
+++ b/daemon/src/chain_client/polygon.rs
@@ -58,14 +58,19 @@ use super::{
     GeneralTransactionId,
     KeyringClient,
     QueryError,
+    SettlementTransfer,
+    SettlementVerificationError,
     SignPermitRequestData,
     SignedTransaction,
     SignedTransactionUtils,
     SubscriptionError,
+    SwapSettlement,
+    SwapSettlementVerifier,
     TransactionError,
     TransfersStream,
     UnsignedTransaction,
 };
+use crate::types::SwapChainType;
 
 use super::keyring::SignTransactionRequestData;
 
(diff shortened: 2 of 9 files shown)


kirushik:
## Blocked: this does not close the hole it targets. Deferring to 0.10.

Two review passes ran against `deeb6c69` — an adversarial code review and an integration review that verified every external assumption against first-hand sources (vendored crate sources for the pinned versions, the Ethereum execution specs, sqlite.org, and live provider APIs). Both are summarised here so the next attempt doesn't have to rediscover any of it.

I've converted this to draft and we're shipping 0.9.5 without it, with a much smaller containment change instead (see the bottom).

### Blocker 1 — the settlement evidence is not bound to the payer

`update_swap_submitted_with_hash` (`daemon/src/dao/swap.rs`) binds a caller-supplied transaction hash to any swap id, with no ownership check and no status precondition. `settlement_matches` (`daemon/src/swaps/tracker.rs`) then asks only whether *some* transfer of the expected token reached the expected address on the expected chain.

So the original attack survives essentially intact. Create a swap with your own `from_address`, then submit **the victim's own genuine payment transaction hash**. That transaction really did move the configured token to the invoice's `payment_address` on Polygon, so the predicate passes, the marker is written, and the attacker's address becomes the refund destination.

The decoded ERC-20 `Transfer` event does carry a `from`, but `SettlementTransfer` (`daemon/src/chain_client.rs`) discards it, so no payer binding exists anywhere in the evidence. Note that for a 0x same-chain swap the token movement is `Settler -> taker`, so the event's `from` is the router contract, not the payer — the binding that would actually work is the **transaction sender** on the receipt, compared against the swap's `from_address`.

### Blocker 2 — the amount is never checked

`settlement_matches` ignores `amount_units` entirely; the aggregate is only logged. A `transfer(payment_address, 0)` of the configured token — a few cents of gas — satisfies verification outright. Values exceeding `u128` decode to `None` and still pass.

Together, these reduce the fix from "any successful Polygon transaction" to "any transaction that moved the right token to this invoice's address, including zero."

The stated reason for making the amount advisory was that no defensible bound could be reconstructed without the quoted slippage and the destination token's decimals. That looks wrong and is worth revisiting: `expected_to_amount_units` is a `u128` in base units and the decoded `amount_units` is too, so they are directly comparable and no decimals are needed.

### Across can be verified — the decision record here is inaccurate

This is the finding that most changes the design. `docs/swaps.md` in this PR states that Across cannot be verified because its stored reference identifies a source-chain deposit. The premise is right; the conclusion is not.

`GET /api/deposit/status` — the endpoint the daemon **already calls** — returns `fillTxnRef` and `destinationChainId` alongside `depositTxnRef`. The daemon **already parses both**, at `daemon/src/clients/swaps/across/types.rs`, behind `#[expect(dead_code)]`, and discards everything but `status`. Recipient and output amount are one further call away via `/api/deposit`.

So making every cross-chain swap permanently refund-ineligible was an unnecessary regression, justified by a limitation that isn't real. The honest caveat to document instead is a different one: following Across's pointer verifies the fill but roots trust in Across's indexer rather than in an independently derived expectation.

### Confirmed sound, so the next attempt can keep it

- **No external API change.** Neither OpenAPI spec in `kalatori-api` mentions swaps or `error_message`; `PublicSwap` omits the field and is the return type of every swap route, so the marker cannot leak. kokpitti's `PublicSwap` model matches field-for-field and konductor doesn't touch swaps.
- **Reverted transactions cannot satisfy the check**, despite `status()` never being consulted — a reverted transaction's receipt carries zero logs by EVM consensus rule (`execution-specs`: on error the log set is replaced wholesale, and `incorporate_child_on_error` doesn't propagate child logs). It's correct, but it rests on an unstated invariant; per this repo's own `#[expect]` convention that deserves either the one-line check or a comment citing the rule.
- **Address handling is right.** alloy's `Display for Address` emits EIP-55 mixed case, so `eq_ignore_ascii_case` is required, not incidental.
- **`log_decode` validates topic0 and topic count**, so an ERC-721 `Transfer` (4 topics) is rejected; the emitting contract isn't validated by the decoder but the explicit `log.address()` comparison covers it.
- **`Failed` is genuinely terminal.** The `BEFORE UPDATE OF status` trigger fires syntactically on any SET touching `status`, `RAISE(ABORT)` reliably backs out the statement while leaving the transaction active, and there is no `INSERT OR REPLACE` anywhere to bypass it.

### Also worth carrying forward

- **`error_message` has no trigger protection.** `Failed` being terminal is what makes the marker unforgeable today, but `update_swap_completed` has no status guard, and any future UPDATE that writes `error_message` without naming `status` wouldn't fire the trigger. The marker is a caller assertion, not a schema invariant — a dedicated column with its own constraint is the right home.
- **`Abandoned` is missing from the transition trigger**, so `Abandoned -> Completed` is currently legal.
- **Native-token settlement produces no ERC-20 `Transfer` at all.** 0x's native sentinel is a supported configuration here (`zeroex/types.rs` maps `0x000…0` to it), and wrapped-token settlement emits `Deposit`/`Withdrawal`, not `Transfer`. Under either, every legitimate swap would fail verification and be marked terminally `Failed`.
- **No confirmation depth on the receipt.** The transfer tracker deliberately buffers 12 confirmations because Polygon reorgs; this reads the receipt at whatever depth it happens to be and a mismatch is terminal, so a reorg can terminally fail a legitimate swap.
- **The chain comparison is tautological today** — both sides are always Polygon — though it fails safe if a second EVM chain is added.
- **Tests don't exercise the security wiring.** The DAO test asserts eligibility can be *set* rather than that verification produced it; the amount-tolerance test calls the predicate directly and would pass even if the real path rejected the settlement; the receipt decoder and the verifier-error paths are untested. Whatever lands next needs a test that reproduces the arbitrary-hash attack end to end.

### What 0.9.5 ships instead

The refund-destination detector will simply stop auto-selecting a destination from swaps. Refunds then resolve to a genuine on-chain payer or fall through to `NoAvailableDestination` and manual handling. It is a few lines, it closes the theft outright, and it depends on none of the analysis above being right — which is the property worth having in a point release.

The proper fix belongs in 0.10, and now has a clear shape: bind the receipt's transaction sender to `from_address`, require the delivered amount to meet `expected_to_amount_units`, verify Across through `fillTxnRef`, give eligibility a dedicated column with trigger protection, honour confirmation depth, and handle the native-token configurations.

